PR #46, opened 2024-03-26
feat: MCR resource

MegaportPhilipBrowne:
## Description

Adds MCR Resource and CRUD methods to Terraform Provider Upgrade


internal/provider/mcr_resource.go
```diff
@@ -0,0 +1,589 @@
...
+	_ resource.ResourceWithConfigure   = &mcrResource{}
+	_ resource.ResourceWithImportState = &mcrResource{}
+)
+
+// mcrResourceModel maps the resource schema data.
+type mcrResourceModel struct {
+	LastUpdated types.String `tfsdk:"last_updated"`
+
+	ID                    types.Int64  `tfsdk:"product_id"`
+	UID                   types.String `tfsdk:"product_uid"`
```
mega-alex, line 32: the port resource uses `uid` instead of `product_uid` we should probably keep them consistent

MegaportPhilipBrowne: Change pushed to have both the Port and MCR resources use `product_uid` to be consistent with the Megaport API.

internal/provider/mcr_resource.go
```diff
@@ -0,0 +1,589 @@
...
+// Schema defines the schema for the resource.
+func (r *mcrResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Megaport Cloud Router (MCR) resource for Megaport Terraform provider.",
+		Attributes: map[string]schema.Attribute{
+			"last_updated": schema.StringAttribute{
+				Description: "Last updated by the Terraform provider.",
+				Computed:    true,
+			},
+			"uid": schema.StringAttribute{
```
mega-alex, line 172: Schema needs to be updated here for the correct uid field

mega-alex (approved):

PR #49, opened 2024-04-08
feat: add vxc resource to upgrade branch

MegaportPhilipBrowne:
feat: vxc resource for upgrade branch


internal/provider/vxc_resource.go
```diff
@@ -0,0 +1,1494 @@
...
+	CreateDate        types.String `tfsdk:"create_date"`
+	ContractStartDate types.String `tfsdk:"contract_start_date"`
+	ContractEndDate   types.String `tfsdk:"contract_end_date"`
+
+	PortUID           types.String              `tfsdk:"port_uid"`
+	AEndConfiguration *vxcEndConfigurationModel `tfsdk:"a_end"`
+	BEndConfiguration *vxcEndConfigurationModel `tfsdk:"b_end"`
+
+	AEndOrderConfiguration *vxcOrderEndPointConfigurationModel `tfsdk:"a_end_order_configuration"`
+	BEndOrderConfiguration *vxcOrderEndPointConfigurationModel `tfsdk:"b_end_order_configuration"`
```
mega-alex, line 60: I would merge these into one type, so that there is only one place to edit fields like `InnerVLAN`. It's a little hard to merge that with the MVE/MCR destination options, but at the very least we shouldn't have an input and an output in the model.

internal/provider/vxc_resource.go
```diff
@@ -0,0 +1,1494 @@
...
+	UID      types.String `tfsdk:"uid"`
+	Type     types.String `tfsdk:"type"`
+	NewSpeed types.Int64  `tfsdk:"new_speed"`
+}
+
+// vxcOrderEndPointConfigurationModel maps the end point configuration schema data.
+type vxcOrderEndPointConfigurationModel struct {
+	ProductUID              types.String             `tfsdk:"product_uid,omitempty"`
+	VLAN                    types.Int64              `tfsdk:"vlan,omitempty"`
+	PartnerConfig           *vxcPartnerConfiguration `tfsdk:"partner_config,omitempty"`
```
mega-alex, line 235: This should probably be modeled where each partner is it's own field rather than all the fields being set in the schema at the root level. Could also be a general purpose config section where it has options like:

```go
type vxcOrderEndPointConfigurationModel struct {
	ProductUID              types.String             `tfsdk:"product_uid,omitempty"`
	VLAN                    types.Int64              `tfsdk:"vlan,omitempty"`
	Config *endConfig
}

type endConfig struct {
	MVEConfig vxcOrderMVEConfigModel
	AWSConfig vxcPartnerConfigAWSModel
	// ... etc
}
```
Then we can enforce that only one is set at a time using terraform validation rules.

PR #60, opened 2024-05-15
Fix/attribute updates

mega-alex:
Fixes `market` and `marketplace_visibility` to better reflect the API. Also fixes a description string for `megaport_port.resources`


mega-alex:
Fixed in #61

PR #137, opened 2024-08-14
fix: change azure service key to 197d927b-90bc-4b1b-bffd-fca17a7ec735

MegaportPhilipBrowne:


mega-alex:
Needs to resolve conflicts, but other than that seems fine

mega-alex (approved):

PR #174, opened 2024-10-18
fix: support import of MVE by only requiring replace for non-imported MVE vendor_config

MegaportPhilipBrowne:


internal/provider/mve_resource.go
```diff
@@ -936,13 +940,19 @@ func (r *mveResource) ImportState(ctx context.Context, req resource.ImportStateR
...
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
 	if !req.State.Raw.IsNull() {
 		stateDiags := req.State.Get(ctx, &state)
 		resp.Diagnostics.Append(stateDiags...)
 	}
 	if resp.Diagnostics.HasError() {
+		fmt.Println("state error!!!")
```
mega-alex, line 955: This looks like debug code that could be a better error / passed through

mega-alex (changes requested):

PR #208, opened 2025-02-06
fix: add rate limit logic for reading/updating/deleting mcr prefix filter lists in provider

MegaportPhilipBrowne:


internal/provider/mcr_resource.go
```diff
@@ -764,21 +808,36 @@ func (r *mcrResource) Read(ctx context.Context, req resource.ReadRequest, resp *
...
+	rateLimiter := NewRateLimiter(10, 1000*time.Millisecond)
+
 	for _, l := range prefixFilterLists {
 		wg.Add(1)
 		go func(list *megaport.PrefixFilterList) {
 			defer wg.Done()
+			// Get a token from the rate limiter to apply rate limiting
+
+			for {
+				if rateLimiter.GetToken() {
```
mega-alex, line 822: Why did you do it this way and not something like 

```go
select {
 case <-rl.rateLimitCh: // in this case we'd need to expose the channel
}
```
this should block until there is a token to be read

internal/provider/mcr_resource.go
```diff
@@ -976,39 +1026,51 @@ func (r *mcrResource) Update(ctx context.Context, req resource.UpdateRequest, re
...
-			"Error Updating Prefix Filter Lists",
-			"Could not update prefix filter lists for MCR with ID "+state.UID.ValueString()+": "+errStr,
+			"Error Modifying Prefix Filter Lists",
+			"Could not modify prefix filter lists for MCR with ID "+state.UID.ValueString()+": "+errStr,
 		)
 		return
 	}
 
 	wg2 := sync.WaitGroup{}
+	mux2 := sync.Mutex{}
```
mega-alex, line 1045: You don't need another mutex, or honestly another wait group.

mega-alex (approved):

PR #266, opened 2025-06-30
feat: support ipMtu in vrouter/aend partner config

MegaportPhilipBrowne:
### User-Facing Summary

This PR adds support for IP MTU (Maximum Transmission Unit) configuration in the Megaport Terraform provider. The IP MTU setting was implemented in the Megaport API in April 2025, but was missing from the Terraform provider. This addition allows users to configure custom MTU values for their virtual cross connects, which is particularly valuable when connecting to cloud providers that support jumbo frames (like Oracle Cloud Infrastructure).

---

### Type of Change

- [x] 🚀 New Feature
- [x] ✨ Enhancement
- [ ] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### Related Issues

---

### How to Test

* Create a new MCR (Megaport Cloud Router) or update an existing one
* Create a VXC (Virtual Cross Connect) between your MCR and a cloud provider that supports jumbo frames
* Configure the `ip_mtu` parameter (either globally or on specific ends) to a value larger than the default 1500
* Apply the configuration and verify that:
1. The resource is created/updated successfully
2. The MTU is set correctly in the Megaport Portal UI
3. The MTU setting is preserved on subsequent terraform plan operations

---

### DIFF
--- a/internal/provider/vxc_resource.go
+++ b/internal/provider/vxc_resource.go
@@ -147,6 +147,7 @@ var (
 
 	// deprecated
 	vxcInterfaceAttrs = map[string]attr.Type{
+		"ip_mtu":           types.Int64Type,
 		"ip_addresses":     types.ListType{}.WithElementType(types.StringType),
 		"ip_routes":        types.ListType{}.WithElementType(types.ObjectType{}.WithAttributeTypes(ipRouteAttrs)),
 		"nat_ip_addresses": types.ListType{}.WithElementType(types.StringType),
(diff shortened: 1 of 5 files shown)


internal/provider/vxc_resource.go
```diff
@@ -147,6 +147,7 @@ var (
 
 	// deprecated
 	vxcInterfaceAttrs = map[string]attr.Type{
```
mega-alex, line 149: we don't need to update the deprecated bits, unless this is causing issues?

mega-alex (approved):

Issue #328, opened 2026-03-19
[BUG] Provider evaluating both branches of a ternary conditional during plan


michaelmyers-payroc:
**Describe the bug**
When using conditional logic to select between MCR UIDs during `terraform plan`, the provider fails with a "Value Conversion Error" even though the conditional should prevent evaluation of the unknown values. The error occurs because Terraform evaluates both branches of a ternary conditional during the plan phase, and the Megaport provider's `vxcEndConfigurationModel` type cannot handle unknown values that would only exist when the condition is true.

**Error Message:**
```
Error: Value Conversion Error

  with module.megaport_vxc_port_vxc_pri["elk"].megaport_vxc.vxc,
An unexpected error was encountered trying to build a value. This is always
an error in the provider. Please report the following to the provider
developer:

Received unknown value, however the target type cannot handle unknown
values. Use the corresponding `types` package type or a custom type that
handles unknown values.

Path: 
Target Type: provider.vxcEndConfigurationModel
Suggested Type: basetypes.ObjectValue
```

**Terraform and Provider Version**
- **Terraform Version:** `1.5.7`
- **Provider Version:** `megaport/megaport ~>1.4.6`

**Terraform File and Steps to Reproduce**
We have a DR (Disaster Recovery) architecture where we want to conditionally point VXCs to either BAU (Business As Usual) MCRs or DR MCRs based on a boolean variable `enable_dr`.

When `enable_dr = false`, only the BAU MCRs exist, and the DR MCRs do not. However, Terraform still validates the conditional expression that references the DR MCR UIDs, causing the provider to receive unknown values.

### Minimal Reproducible Configuration

**variables.tf:**
```hcl
variable "enable_dr" {
  type        = bool
  default     = false
  description = "Enable Disaster Recovery MCRs and VXCs"
}
```

**mcr_dr.tf:**
```hcl
module "megaport_mcr_dr" {
  source   = "../../../../../modules/megaport/mcr"
  for_each = var.enable_dr ? local.mcr_dr_site_data : {}

  # ... MCR configuration ...
}
```

**vxc.locals.tf:**
```hcl
locals {
  # MCR UIDs - conditionally select BAU or DR MCRs
  mcr_uids = var.enable_dr ? {
    us_blue = module.megaport_mcr_dr["dfw3"].megaport_mcr.product_uid
    us_red  = module.megaport_mcr_dr["dfw10"].megaport_mcr.product_uid
  } : {
[...]

PR #329, opened 2026-03-19
ESD-913: fix VXC ModifyPlan crash when end configurations contain unknown values from conditional expressions

Phil-Browne:
### User-Facing Summary

Fixes a `terraform plan` crash ("Value Conversion Error") when a VXC's `a_end_configuration` or `b_end_configuration` is chosen with a conditional expression (for example a ternary for DR failover). Terraform evaluates both branches during plan, so an end config can be wholly unknown while it references a resource that does not exist yet. ModifyPlan now skips reconciliation when an end config is unknown (or null) and lets Terraform re-evaluate once the values resolve.

---

### Type of Change

- [ ] 🚀 New Feature
- [ ] ✨ Enhancement
- [x] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### What changed

- ModifyPlan early-returns when any `a_end`/`b_end` config object is wholly unknown or null, instead of decoding it and crashing.
- An unknown `requested_product_uid` is preserved through reconciliation rather than being clobbered by the cloud-provider branch.
- Pulled the per-end logic into a `reconcileVXCEnd` helper and dropped its dead state writes (ModifyPlan cannot mutate state, so those writes never took effect).

---

### Related Issues

#328, ESD-913

---

### Verification

Done locally so reviewers do not need to re-run:

- `golangci-lint run` clean; full `go test ./internal/provider/` passes.
- Unit tests cover the early-return guard (unknown and null ends), the `requested_product_uid` switch, and the CSP RequiresReplace branch. The behavior-bearing cases are mutation-tested: each fails if the fix is reverted.
- Acceptance tests on staging, all passing:
  - `TestAccMegaportVXC_Basic`
  - `TestAccMegaportVXC_AttachedProductReplace`
  - `TestAccMegaportVXC_ImportDrift_WithPartnerConfig`

---

### How to Test

1. Create a Terraform config with a conditional expression in a VXC's `a_end_configuration` or `b_end_configuration` that references a resource which may not exist (for example `condition ? resource_a.uid : resource_b.uid`).
2. Run `terraform plan`.
3. Confirm the plan completes without a "Value Conversion Error" crash.


internal/provider/vxc_resource.go
```diff
@@ -2815,18 +2815,18 @@ func (r *vxcResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanReq
...
-			// until the upstream resource is applied.
-			if aEndStateObj.IsUnknown() || bEndStateObj.IsUnknown() || aEndPlanObj.IsUnknown() || bEndPlanObj.IsUnknown() {
-				resp.Diagnostics.Append(diags...)
+			// Skip the partner-config / UID reconciliation below if any end-config
+			// object is wholly unknown or null. Unknown objects arise from
+			// conditional expressions referencing resources not yet applied; both
+			// unknown and null objects would fail to decode into a struct and
+			// produce a confusing "Value Conversion Error" rather than a useful
+			// diagnostic. (a_end/b_end are Required, so null is defensive.)
+			if aEndStateObj.IsUnknown() || bEndStateObj.IsUnknown() || aEndPlanObj.IsUnknown() || bEndPlanObj.IsUnknown() ||
```
projectsuperba: Just for readability can these conditions be moved to a small helper function?

projectsuperba (changes requested):

internal/provider/vxc_resource.go
Phil-Browne: Extracted into a small helper, `anyVXCEndObjectUnknownOrNull`, in 070c089.

projectsuperba (approved):

PR #347, opened 2026-04-10
ESD-784(v2): add per-resource configurable timeouts

Phil-Browne:
Replaces the provider's single global `wait_time` (a thread-unsafe package-level `var waitForTime`) with per-resource `timeouts` blocks via `terraform-plugin-framework-timeouts`, giving granular control over provisioning waits per resource and per operation.

The old global was also buggy: the IX resource ignored it (hard-coding `10 * time.Minute`), VXC used a separate `const updateTimeout = 120 * time.Second`, and parallel test/provider instances could corrupt the shared value.

## What changed

**Provider (`provider.go`)**
- Moved `waitForTime` off the package global onto `megaportProviderData`.
- Provider-level `wait_time` is kept as the default fallback.

**All 6 resources** (port, lag_port, mcr, mve, vxc, ix)
- Added a `timeouts` block (create, update, delete) and `Timeouts timeouts.Value` to each model.
- Create/Update resolve the timeout from the block, falling back to the provider `wait_time`.

**Bug fixes**
- IX: replaced hard-coded `10 * time.Minute` with the configurable timeout.
- VXC: removed `const updateTimeout` — the update timeout now flows through `waitForVXCUpdate` and `waitForVnicIndex`.

**Usage**
```hcl
resource "megaport_vxc" "example" {
  # ...
  timeouts {
    create = "20m"
    update = "5m"
    delete = "10m"
  }
}
```

## Not in scope
- `mcr_prefix_filter_list` — no provisioning waits, so timeouts would be schema noise.
- Provider `wait_time` deprecation — kept for backward compatibility.

## Test plan
- [x] `go build -v ./...`
- [x] `go test -v -timeout=30m ./internal/provider/` (existing tests unchanged — default fallback to provider `wait_time`)
- [x] `golangci-lint run --timeout=10m` (0 issues)


Phil-Browne:
Superseded by #416, a clean re-cut of this change against `feat/v2`. This PR was built on the now-abandoned v2 stack.

PR #352, opened 2026-04-14
ESD-793(v2): MVE resource V2 comprehensive cleanup

Phil-Browne:
This PR implements the V2 cleanup of the `megaport_mve` resource (ESD-793), combining noise field removal, vNIC VLAN removal, per-vendor config refactor, and V1→V2 state migration into a single PR.

## What changed

**Noise field removal**

Removed 17 fields that add no practitioner value: `product_id`, `product_type`, `provisioning_status`, `usage_algorithm`, `company_name`, `locked`, `admin_locked`, `cancelable`, `live_date`, `create_date`, `contract_start_date`, `contract_end_date`, `vxc_permitted`, `buyout_port`, `virtual`, `secondary_name`, `last_updated`.

**vNIC VLAN removal**

Removed the `vlan` field from the `vnics` nested attribute. The VLAN is an internal virtual function detail — not useful to customers. The vNIC index is sufficient for ordering.

**Per-vendor config blocks**

Replaced the single `vendor_config` union block (26 fields, all optional) with 10 separate top-level optional blocks: `aruba_config`, `aviatrix_config`, `cisco_config`, `fortinet_config`, `meraki_config`, `palo_alto_config`, `prisma_config`, `sixwind_config`, `versa_config`, `vmware_config`. Each block contains only the fields relevant to that vendor. An `objectvalidator.ExactlyOneOf` validator ensures exactly one block is set. Sensitive fields (account keys, tokens, activation codes, passwords) are marked `Sensitive: true`. The top-level `vendor` attribute remains as a computed field populated by the API.

**V1→V2 state migration**

Implements `ResourceWithMoveState` so that running `terraform state mv` from a V1 provider to V2 automatically migrates existing `megaport_mve` state. The migration:
- Drops all 17 removed fields silently
- Strips `vlan` from each vNIC
- Maps the V1 `vendor_config` union block to the correct V2 per-vendor config block based on the `vendor` field
- Handles imported resources (null vendor_config) gracefully by leaving all config blocks null

## Import behavior

After importing an MVE, all vendor config blocks will be null (the API does not return vendor config on read). Add `lifecycle { ignore_changes = [aruba_config] }` (substituting your vendor) to prevent Terraform from attempting to recreate the resource.

## Testing

15 new unit tests in `mve_move_state_test.go` covering all 10 vendor migrations (Aruba, Aviatrix, Cisco, Fortinet, Meraki, Palo Alto, Prisma, Sixwind, Versa, VMware), null vendor_config handling, vNIC VLAN stripping, nil optionals, wrong provider, and wrong type.


Phil-Browne:
Closing as part of retiring the stacked v2 branches (this one sat 4 deep, on the branch of the now-closed #347). The MVE field deprecations are re-cut as #411 against the synced feat/v2. The per-vendor config split is being redesigned before re-cutting: this branch forked before the Cisco FTDv admin_password (write-only) field landed on main, so merging it as-is would silently drop that field.

PR #357, opened 2026-04-16
ESD-914: add megaport_vxcs data source

Phil-Browne:
### User-Facing Summary

Adds a new `megaport_vxcs` data source following the same pattern as the existing `megaport_mcrs` data source. Supports listing all active VXCs or looking up a specific VXC by `product_uid`.

Exposed attributes include product name, rate limit, provisioning status, A-End/B-End configuration (UID, name, location ID, VLAN), contract details, distance band, and both attribute and resource tags. Resource tag fetching is opt-in via `include_resource_tags` to avoid N+1 API calls.

Also bumps `megaportgo` from v1.4.9 to v1.7.0 and fixes the pre-existing MCR mock compilation error (missing `UpdateMCRIPsecAddOn`/`UpdateMCRWithAddOn` interface stubs).

---

### Type of Change

- [x] 🚀 New Feature
- [ ] ✨ Enhancement
- [x] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [x] 🏗️ Chore / Other

---

### Related Issues

Supersedes #244 (VXC data source portion only)

---

### How to Test

Unit tests:
```bash
go test -v -timeout=5m -run "TestReadVXCs|TestFromAPIVXCDetail|TestVXCsModel" ./internal/provider/
```

Full build + lint:
```bash
go build -v ./...
golangci-lint run --timeout=10m
go generate ./...  # verify no diff
```

---


mp1-tomclancy (approved):
LGTM.

It seems like it's common for TF providers to provide plural/singular data source counterparts e.g. [github_repository](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/repository) vs. [github_repositories](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/repositories). 

As a future enhancement, we could do the same. It would provide a better developer experience when fetching a single resource. If we choose this approach then we should also apply to other data sources (e.g. MCR).

PR #359, opened 2026-04-16
ESD-916: Add megaport_ports data source

Phil-Browne:
Add the `megaport_ports` data source, following `megaport_mcrs` and `megaport_vxcs`.

Set `product_uid` to fetch one port, or omit it to list all active ports.

**Changes**

- Add the data source schema, read logic, and unit tests.
- Return a port by UID regardless of provisioning status.
- Add `include_resource_tags` to opt in to tag fetches.
	- Default off. Tags cost one API call per port.
- Sort list output by product UID.
- Leave out `attribute_tags`.
	- `Port.AttributeTags` is a struct, not a string map.
- Move `MockVXCService` into its own shared file.
- Generate `docs/data-sources/ports.md`.

**Testing**

Build, lint, unit tests, and the Generate gate pass. No acceptance test. Harness run on the current head: [35132913179](https://github.com/megaport/terraform-provider-acceptance-ci/actions/runs/35132913179).

**Scope of change**

Customer-facing yes, new data source. Existing automated coverage no, unit tests only.


internal/provider/ports_data_source.go
```diff
@@ -0,0 +1,406 @@
...
+		if port == nil {
+			resp.Diagnostics.AddError(
+				"Error reading port",
+				"Port not found: "+data.ProductUID.ValueString(),
+			)
+			return
+		}
+		ports = []*megaport.Port{port}
+	} else {
+		// List all ports
```
projectsuperba: Should there be a check for `DECOMMISSIONED` and `CANCELLED` port status to exclude those here and for the single port case above?

Phil-Browne:
Good catch @projectsuperba  — pushed `effb637` addressing this.

`megaportgo`'s `PortService.ListPorts` doesn't take a status filter (unlike `ListMCRs`/`ListMVEs`/`ListVXCs` which accept `IncludeInactive`), so the data source was returning every port regardless of status despite the schema promising active ports only. I've added client-side filtering:

- **list case** — skip ports whose `ProvisioningStatus` is `DECOMMISSIONED` or `CANCELLED`
- **single-UID case** — same statuses are treated as not-found, with the existing "Port not found" diagnostic

Added `TestReadPorts_ListSkipsInactive` and `TestReadPorts_GetByUIDRejectsInactive` to cover both paths.

internal/provider/ports_data_source.go
```diff
@@ -0,0 +1,421 @@
...
+		Locked:                types.BoolValue(p.Locked),
+		AdminLocked:           types.BoolValue(p.AdminLocked),
+		Cancelable:            types.BoolValue(p.Cancelable),
+		DiversityZone:         types.StringValue(p.DiversityZone),
+	}
+
+	// Time fields — emit RFC3339 so values are consumable by Terraform's
+	// formatdate() function; nil dates map to null rather than an empty string.
+	if p.CreateDate != nil {
+		detail.CreateDate = types.StringValue(p.CreateDate.Format(time.RFC3339))
```
projectsuperba, line 432: I have a broader question about time formats in this/other repos. RFC3339 looks like the right standard for Terraform dates. But this repo has three formats today:

- **RFC850**: most resources, e.g. [single_port_resource.go:106](https://github.com/megaport/terraform-provider-megaport/blob/effb637ace59c4fcedcd97547d0499a1a1c3a8db/internal/provider/single_port_resource.go#L106)
- **RFC3339**: this PR's data source shown above, plus [ix_resource.go:179](https://github.com/megaport/terraform-provider-megaport/blob/effb637ace59c4fcedcd97547d0499a1a1c3a8db/internal/provider/ix_resource.go#L179)
- **Go's default `time.Time.String()`**: [mcrs_data_source.go:347](https://github.com/megaport/terraform-provider-megaport/blob/effb637ace59c4fcedcd97547d0499a1a1c3a8db/internal/provider/mcrs_data_source.go#L347)

If there's no reason to keep RFC850, we should standardize the repo on RFC3339. Users whose state currently holds RFC850 strings would see a one-time "changed outside of Terraform" notice on next plan and the state auto-updates on the next `apply`. Any poweruser scripting against the text output would break on that one format change but that's fair game for a minor-version bump with changelog notes though. This all assumes anyone has been using this version/s of code externally, I'm not familiar yet with this release/endeavor.

@ianb-mp opinion on which path?

1. Just find and replace all dates to be RFC3339 and accept the one-time drift messages. (@projectsuperba votes this)
2. Add a migration helper in each resource's `Read()` to rewrite RFC850 → RFC3339 on the fly, so users don't see the drift notice (requires followup removal of code later).
3. Leave resources on RFC850 and align data sources to match (seems... incorrect to do).

projectsuperba (commented):
Perhaps as a followup - new acceptance test for the new data source?

PR #360, opened 2026-04-16
ESD-917: add megaport_mves data source

Phil-Browne:
### User-Facing Summary

Adds a new `megaport_mves` data source that allows users to look up MVE (Megaport Virtual Edge) resources from the Megaport API. This follows the same pattern established by the `megaport_mcrs` and `megaport_ixs` data sources:

- Look up all active MVEs, or a specific MVE by `product_uid`
- Opt-in `include_resource_tags` to avoid N+1 API calls by default
- Exposes key MVE fields: vendor, size, diversity_zone, provisioning_status, contract details, attribute_tags, resource_tags, and more (a curated subset suitable for listing; the full MVE resource exposes additional fields like vendor_config and vnics)

Also bumps `github.com/megaport/megaportgo` from v1.4.9 to v1.7.0, which is required for the new `MVEService.ListMVEs` and `MVEService.ListMVEResourceTags` methods used by this data source. Test mocks are updated accordingly to satisfy the new MCR service interface.

#### `megaport_mcrs` data source behavior change

The MCRs data source time formatting was aligned to use RFC850 and `null` for nil dates, replacing the prior empty-string (`""`) representation. This is intentional — all V2 data sources now consistently return `null` for unset timestamps rather than empty strings. Configurations that check `== ""` on MCR date fields will need to use `== null` instead.

---

### Type of Change

- [x] 🚀 New Feature
- [ ] ✨ Enhancement
- [x] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### Related Issues

---

### How to Test

\`\`\`bash
# Build
GOWORK=off go build -v ./...

# Run unit tests
GOWORK=off go test -v -timeout=5m -run "TestReadMVEs|TestFromAPIMVEDetail|TestMVEsModel|TestFromAPIMCRDetail" ./internal/provider/

# Lint
GOWORK=off golangci-lint run --timeout=10m
\`\`\`

PR #364, opened 2026-04-21
ESD-918: drive mcrs Read end-to-end and guard nil GetMCR result

Phil-Browne:
The `TestReadMCRs_*` tests exercised `MockMCRService` directly instead of invoking the data source's `Read` method, so the wiring between config, service dispatch, and state was never tested. This rewrites them to build a real `datasource.ReadRequest` and call `ds.Read(ctx, req, resp)`, asserting on `resp.State` and diagnostics. It follows the pattern already used by the ixs/mves/ports/vxcs data source tests (Stuart's review feedback on #360).

Driving `Read` end-to-end surfaced a latent nil dereference: `GetMCR` can return `(nil, nil)` when the API resolves a UID to a null body, which would panic in the detail loop. It now returns an error diagnostic instead.

## Changes

- Rewrite `TestReadMCRs_*` to call `ds.Read` end-to-end via a schema-built `ReadRequest`.
- Guard `(nil, nil)` from `GetMCR` in `Read` with an error diagnostic.
- Add the `DataSourceWithConfigure` compile-time interface guard to match the peer data sources.
- New coverage: per-MCR tag fetch and warning fallback, empty list, nil result, RFC3339 time formatting, and attribute/resource tag mapping.

## Test plan

- [x] `go test ./internal/provider/` (full unit suite passes)
- [x] `go build ./...`, `go vet ./internal/provider/`
- [x] `golangci-lint run --timeout=10m ./internal/provider/` (0 issues)
- [x] `go generate ./...` unaffected (no schema change)

### DIFF
--- a/internal/provider/mcrs_data_source.go
+++ b/internal/provider/mcrs_data_source.go
@@ -14,7 +14,8 @@ import (
 
 // Ensure the implementation satisfies the expected interfaces
 var (
-	_ datasource.DataSource = &mcrsDataSource{}
+	_ datasource.DataSource              = &mcrsDataSource{}
+	_ datasource.DataSourceWithConfigure = &mcrsDataSource{}
 
 	mcrDetailAttrs = map[string]attr.Type{
 		"product_uid":            types.StringType,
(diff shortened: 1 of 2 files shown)


Phil-Browne:
@copilot resolve the merge conflicts in this pull request

projectsuperba (approved):

PR #368, opened 2026-05-05
ESD-1080: Add admin_password attribute for Cisco FTDv and Palo Alto MVE vendor configs

Phil-Browne:
## Summary

Adds an `admin_password` attribute to the `megaport_mve` resource's `vendor_config` block. The Megaport API requires a plaintext `adminPassword` field on MVE creation for Cisco FTDv and Palo Alto vendors.

## Changes

- `internal/provider/mve_resource.go`
  - Add `admin_password` to the `vendorConfigModel` struct.
  - Wire the value through `toAPIVendorConfig` for the `cisco` and `palo_alto` branches into `megaportgo.CiscoConfig.AdminPassword` and `megaportgo.PaloAltoConfig.AdminPassword`.
  - Add the `admin_password` schema attribute on `vendor_config`. Marked `Sensitive: true` so the value is redacted from plan output and state JSON.
- `docs/resources/mve.md` — regenerated via `tfplugindocs`; the new attribute renders as `(String, Sensitive)`.
- `internal/provider/mve_resource_test.go` — adds `TestAccMegaportMVECisco_Basic` and `TestAccMegaportMVEPaloAlto_Basic` (create + import + update steps).
- `internal/provider/test_helpers_test.go` — adds `findMVECiscoTestLocation` and `findMVEPaloAltoTestLocation` helpers that resolve image IDs dynamically via `ListMVEImages`.

The existing `admin_password_hash` attribute on `palo_alto` configs is preserved for callers that use the hashed credential path.

## Acceptance test notes

- `TestAccMegaportMVECisco_Basic`: self-contained; uses `manage_locally=true` (no FMC required), hardcoded test admin password.
- `TestAccMegaportMVEPaloAlto_Basic`: requires `MEGAPORT_TEST_SSH_PUBLIC_KEY` (RSA 2048 OpenSSH public key) — skips when absent. Uses the documented sha256crypt hash example for `admin_password_hash`.

## Dependency on ESD-1079

This PR depends on [megaportgo PR #153](https://github.com/megaport/megaportgo/pull/153) — `megaportgo.CiscoConfig` and `megaportgo.PaloAltoConfig` both need an `AdminPassword string` field with the `json:"adminPassword,omitempty"` tag.

**This PR cannot merge until:**
1. ESD-1079 lands in `megaportgo` `main`.
2. A new `megaportgo` release is cut containing the new fields.
3. This branch is rebased and `go.mod` is bumped to the released `megaportgo` version.

`go.mod` is not bumped here because the upstream release does not yet exist. Local validation was performed via the parent Go workspace against the in-progress `megaportgo` branch.

## Test plan

- [x] `go build -v ./...` — clean (via Go workspace against local `megaportgo` branch)
- [x] `go test ./internal/provider/` — pass (unit tests)
- [x] `go generate ./...` — docs regenerated
- [x] `gofmt -w .` — clean
- [x] `golangci-lint run` — 0 issues
- [ ] `TestAccMegaportMVECisco_Basic` — run against staging once `megaportgo` release is cut and `go.mod` is bumped
- [ ] `TestAccMegaportMVEPaloAlto_Basic` — run against staging with `MEGAPORT_TEST_SSH_PUBLIC_KEY` set

## Risk

Low. New field is `Optional`, additive on the schema, and write-only at create time. Existing MVE configurations are unaffected.


internal/provider/test_helpers_test.go
```diff
@@ -282,6 +285,139 @@ func findMVEVersaTestLocation(t *testing.T) (id int, name string) {
...
 }
 
+// mveTestPaloAltoAdminPasswordHash is a sha256crypt-formatted password hash used
+// for Palo Alto MVE acceptance tests. Corresponds to the example in the Megaport
+// provider docs. Generated with: mkpasswd -m sha-256 'your_password'
+const mveTestPaloAltoAdminPasswordHash = "$5$2833ea35$Pdyc6dKE8N/UBRge3QWDJJyotG3I59pxLJWVmcSQDdC"
+
+// mveTestCiscoAdminPassword is the plaintext admin password used for Cisco FTDv
+// MVE acceptance tests. Meets API requirements: 9-100 chars, no " CR LF.
+const mveTestCiscoAdminPassword = "TestAdminPw1234!"
```
mp1-tomclancy: These values should be randomly generated during the execution of the acceptance test, given that they are used to create real (temporary) resources.

internal/provider/mve_resource.go
```diff
@@ -642,6 +645,11 @@ func (r *mveResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *
 						Description: `The admin password hash for the vendor config. Required for Palo Alto MVE. Must be a SHA-256 crypt hash in the format "$5$<salt>$<hash>" (e.g., "$5$2833ea35$Pdyc6dKE8N/UBRge3QWDJJyotG3I59pxLJWVmcSQDdC"). On Linux/macOS, you can generate this using: "mkpasswd -m sha-256 'your_password'".`,
 						Optional:    true,
 					},
+					"admin_password": schema.StringAttribute{
+						Description: "The admin password for the vendor config. Required for Cisco FTDv (Firewall) and Palo Alto MVEs.",
+						Optional:    true,
+						Sensitive:   true,
+					},
```
mp1-tomclancy, line 652: `Sensitive` is good, but this doesn't prevent the value from being persisted in state. Consider using [Write-only Arguments](https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments) here. 

I also assume that these credentials are just used for the initial setup of the device, and that users set their own passwords (via vendor-specific management interfaces) after deploying the MVE. If so, maybe we could make this advice explicit in the docs?

docs/resources/mve.md
```diff
@@ -225,6 +225,7 @@ Optional:
 
 - `account_key` (String) The account key for the vendor config. Enter the Account Key from Aruba Orchestrator. The key is linked to the Account Name. Required for Aruba MVE.
 - `account_name` (String) The account name for the vendor config. Enter the Account Name from Aruba Orchestrator. To view your Account Name, log in to Orchestrator and choose Orchestrator > Licensing | Cloud Portal. Required for Aruba MVE.
+- `admin_password` (String, Sensitive) The admin password for the vendor config. Required for Cisco FTDv (Firewall) and Palo Alto MVEs.
 - `admin_password_hash` (String) The admin password hash for the vendor config. Required for Palo Alto MVE. Must be a SHA-256 crypt hash in the format "$5$<salt>$<hash>" (e.g., "$5$2833ea35$Pdyc6dKE8N/UBRge3QWDJJyotG3I59pxLJWVmcSQDdC"). On Linux/macOS, you can generate this using: "mkpasswd -m sha-256 'your_password'".
```
mp1-tomclancy: Does this mean Palo Alto MVEs require *both* `admin_password` and `admin_password_hash`?

mp1-tomclancy (changes requested):

internal/provider/test_helpers_test.go
Phil-Browne: Good call — done in 331ab7a. Replaced the constants with `crypto/rand`-backed generator functions in `test_helpers_test.go`:

- `mveTestCiscoAdminPassword(t)` → 24 chars from a safe alphabet (matches the FTDv 9–100 char rule, avoids `"`, CR, LF per `MveProductValidator.validateFtdv`).
- `mveTestPaloAltoAdminPasswordHash(t)` → `$5$<16-char salt>$<43-char encrypted>` matching `HashAdminPasswordValidator`'s format check.

Each test now generates once at the top and reuses the value across create + update steps so the second apply doesn't try to replace `vendor_config`.

internal/provider/mve_resource.go
Phil-Browne: Done in 331ab7a. `admin_password` is now `WriteOnly: true` (kept `Sensitive: true` as well). Since write-only attributes are null in `req.Plan` / `req.State`, `Create()` now pulls the value from `req.Config` before mapping to the API request. Requires Terraform 1.11+ at apply time; the framework version we're on (v1.18.0) already supports the schema flag.

Also reworded both `admin_password` and `admin_password_hash` descriptions to make the per-vendor scope explicit and to note they are only used for initial device setup — operators are expected to rotate via the device's management interface after deployment.

docs/resources/mve.md
Phil-Browne: No — verified against the OpenAPI spec (`PaloAltoVm.yaml`) and the megalith backend. Palo Alto VM-Series only accepts `adminPasswordHash`; `ProvPaloVmConfigDto` doesn't forward `adminPassword` at all, so it was silently dropped. Cisco FTDv is the inverse — it only accepts `adminPassword`.

Fixed in 331ab7a:
- Removed the `AdminPassword` wiring from the `palo_alto` branch in `toAPIVendorConfig`.
- Updated the schema descriptions so `admin_password` reads as Cisco FTDv only and `admin_password_hash` reads as Palo Alto VM-Series only, with the setup-only caveat on both.
- Regenerated `docs/resources/mve.md`.

mp1-tomclancy (approved):

Issue #369, opened 2026-05-05
[BUG] Changing BGP ASN on an MCR Resource forces replacement


michaelmyers-payroc:
**Describe the bug**
Changing BGP ASN on an MCR Resource forces replacement.

**Terraform and Provider Version**
```
terraform {
  required_providers {
    megaport = {
      source  = "megaport/megaport"
      version = "=1.7.0"
    }
  }
  required_version = ">= 1.5"
}
```

**Terraform File and Steps to Reproduce**
Original:
```
locals {
  mcr_site_data = {
    fr5 = {
      mcr = {
        asn                  = 4200000000
        contract_term_months = 36
        diversity_zone       = "blue"
        location_id          = data.megaport_location.fr5.id
        port_speed           = 10000
        product_name         = "fr5entmcrblu01"
        promo_code           = ""
      }
    },
    fra13 = {
      mcr = {
        asn                  = 4200000000
        contract_term_months = 36
        diversity_zone       = "red"
        location_id          = data.megaport_location.fra13.id
        port_speed           = 10000
        product_name         = "fra13entmcrred01"
        promo_code           = ""
      }
    }
  }
}
```

New ASN:
```
locals {
  mcr_site_data = {
    fr5 = {
      mcr = {
        asn                  = 4200000099
        contract_term_months = 36
        diversity_zone       = "blue"
        location_id          = data.megaport_location.fr5.id
        port_speed           = 10000
        product_name         = "fr5entmcrblu01"
        promo_code           = ""
      }
    },
    fra13 = {
      mcr = {
        asn                  = 4200000099
        contract_term_months = 36
        diversity_zone       = "red"
        location_id          = data.megaport_location.fra13.id
        port_speed           = 10000
        product_name         = "fra13entmcrred01"
        promo_code           = ""
      }
    }
  }
}
```

Module code:
```
locals {
  resource_tags = {
    owner          = "pyrc"
    diversity_zone = var.diversity_zone
    contract_term  = tostring(var.contract_term_months)
    bandwidth      = tostring(var.port_speed)
  }
}

resource "megaport_mcr" "mcr" {
  asn                  = var.asn
  contract_term_months = var.contract_term_months
  diversity_zone       = var.diversity_zone
  location_id          = var.location_id
  port_speed           = var.port_speed
  product_name         = var.product_name
  promo_code           = var.promo_code
  resource_tags        = local.resource_tags
}
```

**Additional context**
[...]

PR #370, opened 2026-05-05
ESD-1095: guard VXC ModifyPlan against unknown end-config objects

Phil-Browne:
## Summary

When a Port/MCR/MVE attached to a VXC is replaced — for example when an MCR's ASN is changed (ESD-1094), or when a Port's `port_speed` (a `RequiresReplace` attribute) is bumped — the upstream resource's `product_uid` is unknown at plan time. That unknown propagates into the dependent VXC's `a_end` / `b_end` `types.Object`, and decoding an unknown object into `vxcEndConfigurationModel` blows up with:

> Value Conversion Error / Target Type: provider.vxcEndConfigurationModel / Suggested Type: basetypes.ObjectValue

This made any plan that touched an attached upstream resource crash before reaching apply, even when the user's intended change was perfectly reasonable.

## Fix

In `vxc_resource.go` `ModifyPlan`:

- Short-circuit the partner-config / `requested_product_uid` reconciliation block when any of the four end-config objects (`state.AEndConfiguration`, `state.BEndConfiguration`, `plan.AEndConfiguration`, `plan.BEndConfiguration`) is `IsUnknown()`. There is nothing meaningful to reconcile until the upstream apply has produced concrete values.
- Guard the `plan.AEndPartnerConfig.As(...)` and `plan.BEndPartnerConfig.As(...)` calls against unknown values for the same reason.

The reconciliation that *was* skipped is purely a plan-time cosmetic — Create/Update still run and produce the correct state.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/provider/` (unit mode)
- [x] `golangci-lint run --timeout=10m` — 0 issues
- [x] `go generate ./...` — no doc drift
- [ ] Acceptance: `TestAccMegaportVXC_AttachedProductReplace` — forces a Port replacement (`port_speed` 1000 → 10000) and asserts the dependent VXC plans cleanly. Pre-fix this step fails with the Value Conversion Error.

## Related

- ESD-1094 (separate PR): MCR ASN should not force replacement. That fix is blocked on a megaportgo SDK change to expose `asn` on `ModifyMCRRequest`. This PR fixes the user-visible cascade error independently, since the same crash can be triggered by any RequiresReplace change on an attached Port/MCR/MVE.

Addresses https://github.com/megaport/terraform-provider-megaport/issues/369


internal/provider/vxc_resource.go
mp1-tomclancy: nit (unrelated): readability in `ModifyPlan` could be improved with helper functions, particularly in cases where the same operations are applied to both ends

mp1-tomclancy (dismissed):

internal/provider/vxc_resource.go
Phil-Browne: Addressed in https://github.com/megaport/terraform-provider-megaport/pull/370/commits/e21eff514aa4e24859d403cdc3b04c2fc803dd97

mp1-tomclancy (approved):

PR #371, opened 2026-05-05
ESD-1094: support modifying MCR ASN in place

Phil-Browne:
## Summary

Changing `asn` on a `megaport_mcr` resource currently forces a destroy-and-recreate, which is disruptive (cancels and re-orders the MCR, and breaks every attached VXC) and unnecessary — the Megaport API has always supported modifying ASN in place via `PUT /v2/product/mcr2/{productUid}`. The provider was forcing replacement only because `megaportgo` did not expose the field.

Depends on Megaportgo PR: [megaport/megaportgo#154](https://github.com/megaport/megaportgo/pull/154) (ESD-1096) adds `ModifyMCRRequest.MCRAsn`.

## Changes

- **Schema (`internal/provider/mcr_resource.go`)**: drop `RequiresReplace()` from the `asn` plan modifier; replace with `UseStateForUnknown()` so plan-time computed handling remains correct. Update the description to call out in-place behavior and the BGP-renegotiation side-effect users should expect.
- **Update path**: when the planned ASN is known and differs from state, pass it into `ModifyMCR` via the new `MCRAsn` field.
- **Dependency**: bump `github.com/megaport/megaportgo` to a build that exposes `MCRAsn`. Final PR will retarget to a tagged release once megaportgo#154 merges.
- **Doc regen** for `docs/resources/mcr.md`.

## Test plan

- [x] `go build ./...` / `go vet ./...`
- [x] `golangci-lint run --timeout=10m` — 0 issues
- [x] `go generate ./...` — committed doc diff
- [x] `go test ./internal/provider/` (unit / `TF_ACC` skip)
- [ ] Acceptance: `TestAccMegaportMCR_UpdateASN` — creates an MCR with asn 64512, updates to 64513, asserts `product_uid` is unchanged (no replacement) and the new ASN reads back. Final plan-only step asserts no drift.

## Related

Addresses https://github.com/megaport/terraform-provider-megaport/issues/369


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/hashicorp/terraform-plugin-framework-validators v0.19.0
 	github.com/hashicorp/terraform-plugin-go v0.30.0
 	github.com/hashicorp/terraform-plugin-log v0.10.0
-	github.com/megaport/megaportgo v1.10.1
+	github.com/megaport/megaportgo v1.10.2-0.20260505152722-0d14697dce02
```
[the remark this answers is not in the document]
Phil-Browne: Agreed — the pseudo-version is intentional for now. The `MCRAsn` field in `ModifyMCRRequest` lives in megaport/megaportgo PR ESD-1096, which is still in review and hasn't been merged to main yet, so there's no `v1.10.2` tag to point at. Once that PR merges I'll tag `v1.10.2` in megaportgo and update this dependency to the released version before merging here.

go.sum
mp1-tomclancy, line 1: Run `go mod tidy` to remove references to older versions of megaportgo

internal/provider/mcr_resource.go
```diff
@@ -685,7 +686,7 @@ func (r *mcrResource) Create(ctx context.Context, req resource.CreateRequest, re
 		WaitForTime:      waitForTime,
 	}
 
-	if !plan.ASN.IsNull() {
+	if !plan.ASN.IsNull() && !plan.ASN.IsUnknown() {
```
mp1-tomclancy: Per [docs](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/accessing-values#when-can-a-value-be-unknown-or-null): 

> Computed attributes, whether optional or not, will never be null in the plan for Create, Read, Update, or Delete methods.

So perhaps the `isNull()` check can be removed

internal/provider/mcr_resource.go
```diff
@@ -981,12 +982,19 @@ func (r *mcrResource) Update(ctx context.Context, req resource.UpdateRequest, re
 		contractTermMonths = &months
 	}
 
+	var mcrAsn *int
+	if !plan.ASN.IsNull() && !plan.ASN.IsUnknown() && !plan.ASN.Equal(state.ASN) {
```
mp1-tomclancy: See my comment on the `Create()` method

internal/provider/mcr_resource.go
```diff
@@ -1117,49 +1125,47 @@ func (r *mcrResource) Update(ctx context.Context, req resource.UpdateRequest, re
...
+	deleteRateLimiter := NewRateLimiter(10, 1000*time.Millisecond)
 
-		for _, stateModel := range statePrefixFilterLists {
-			wg.Add(1)
-			go func(stateModel *mcrPrefixFilterListModel) {
-				defer wg.Done()
+	for _, stateModel := range statePrefixFilterLists {
+		wg.Add(1)
+		go func(stateModel *mcrPrefixFilterListModel) {
+			defer wg.Done()
```
mp1-tomclancy: FYI: `WaitGroup.Go()` is nice for cases like this (but requires `go >= 1.25`)

internal/provider/mcr_resource.go
```diff
@@ -616,6 +616,7 @@ func (r *mcrResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *
 							Computed:    true,
 							PlanModifiers: []planmodifier.Int64{
 								int64planmodifier.UseStateForUnknown(),
+								unknownWhenNoPriorState(),
```
mp1-tomclancy: Would [UseNonNullStateForUnknown](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier#UseNonNullStateForUnknown) work here? 

Haven't tested myself yet 🤔

internal/provider/mcr_resource.go
Phil-Browne: Added https://github.com/megaport/terraform-provider-megaport/pull/371/commits/e588f0370490ae4bbeb1e63429e33aff3aed1290

Phil-Browne:
@mp1-tomclancy Addressed comments in follow up commits

mp1-tomclancy (approved):

PR #385, opened 2026-05-21
ESD-1183: Remove Terminate Later option for Ports

Phil-Browne:
## Summary

Remove the "Terminate Later" (delayed cancellation) option for Single Ports and LAG Ports, mirroring [megaportgo PR #155](https://github.com/megaport/megaportgo/pull/155). The Megaport API no longer accepts delayed cancellation for ports — only immediate cancellation (`CANCEL_NOW`) is allowed.

### Code changes

- Remove the provider-level `cancel_at_end_of_term` configuration attribute (it was previously documented as applying only to Single Ports and LAG Ports).
- Hard-code `DeleteNow: true` in `portResource.Delete` (`internal/provider/single_port_resource.go`) and `lagPortResource.Delete` (`internal/provider/lag_port_resource.go`).
- Drop the now-dead `cancelAtEndOfTerm` field from `megaportProviderData`, `portResource`, and `lagPortResource`.
- Replace the "End-of-Term Cancellation" section in `README.md`, `docs/index.md`, and `templates/index.md.tmpl` with a brief note that cancellation is always immediate.
- Add a unit test (`internal/provider/port_resource_delete_test.go`) asserting the SDK contract used by both port resources hits `/v3/product/{id}/action/CANCEL_NOW`.

## Breaking change

This is a **breaking change for HCL configs** that previously set `cancel_at_end_of_term = true` on the `provider "megaport"` block — Terraform will reject the unknown attribute. Users will need to remove that line from their configuration. There is no behaviour change for users who left the attribute at its default (`false`).

The `RestorePort` SDK method remains in megaportgo for any ports already scheduled for cancellation under the old behaviour; nothing in this provider PR depends on that.

## Dependencies

- **Depends on [megaportgo PR #155](https://github.com/megaport/megaportgo/pull/155)** merging and a new tagged release being published.
- **Requires a `go.mod` version bump for `github.com/megaport/megaportgo`** — this will be added as a follow-up commit on this PR once the new megaportgo tag is available. Do not merge before that bump lands.

## Test plan

- [x] `GOWORK=off go build -v ./...` — clean
- [x] `GOWORK=off go vet ./...` — clean
- [x] `GOWORK=off go test ./internal/provider/... -run '^Test[^A].*'` — passes, including new `TestPortDeleteAlwaysUsesCancelNow`
- [x] `golangci-lint run ./...` — 0 issues
- [x] Confirm `docs/index.md` and `templates/index.md.tmpl` stay in sync (`diff` returns no differences)
- [ ] Acceptance tests (`TF_ACC=1`) — to be run after the megaportgo bump lands on this branch
- [ ] Manual smoke test: apply + destroy a `megaport_port` resource against staging after the SDK bump

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,29 +577,11 @@ data "megaport_partner" "awshc" {
 }
 ```
 
-## End-of-Term Cancellation
+## Resource Cancellation
 
-By default, when Terraform deletes resources, they are immediately cancelled in the Megaport portal. However, you may prefer to have resources marked for cancellation at the end of their current billing term instead of immediate cancellation.
+When Terraform deletes a Megaport resource, the provider issues an immediate cancellation (`CANCEL_NOW`) to the Megaport API. Resources are removed from Terraform state as soon as the API call returns successfully.
 
-The provider supports this with the `cancel_at_end_of_term` configuration option:
-
-```terraform
-provider "megaport" {
-  environment           = "production"
-  access_key            = "your-access-key"
-  secret_key            = "your-secret-key"
-  accept_purchase_terms = true
-  cancel_at_end_of_term = true  # Mark resources for end-of-term cancellation
-}
-```
-
-**Important notes:**
-
-- This feature is currently only supported for Single Ports and LAG Ports
-- For other resource types, the option will be ignored and immediate cancellation will occur
-- When `cancel_at_end_of_term` is set to `true`, resources will show as "CANCELLING" in the Megaport portal until the end of their billing term
-- Resources are removed from Terraform state as soon as the API call returns successfully, regardless of whether immediate or end-of-term cancellation is used
-- If you reapply your configuration after a resource has been deleted, Terraform will create a new resource, even if the original resource is still visible in the Megaport portal with "CANCELLING" status
+Delayed cancellation (cancel-at-end-of-term) is not supported by the provider. The previously available `cancel_at_end_of_term` provider option has been removed because the Megaport API no longer accepts delayed cancellation for Ports or LAG Ports.
 
 ## Importing Existing Resources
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -577,29 +577,13 @@ data "megaport_partner" "awshc" {
 }
 ```
 
-## End-of-Term Cancellation
+## Resource Cancellation
 
-By default, when Terraform deletes resources, they are immediately cancelled in the Megaport portal. However, you may prefer to have resources marked for cancellation at the end of their current billing term instead of immediate cancellation.
+When Terraform deletes a Megaport resource, the provider issues an immediate cancellation or deletion request to the Megaport API. Resources are removed from Terraform state as soon as the API call returns successfully. For Ports and LAG Ports specifically, this is always a `CANCEL_NOW` action against the Megaport Products API.
```
mp1-tomclancy, line 582: Maybe a quick warning that there may be billing implications for the customer, when they delete a resource before the end of its term?

mp1-tomclancy (dismissed):

README.md
Phil-Browne: Addressed in https://github.com/megaport/terraform-provider-megaport/pull/385/commits/e5d7b91c11a78ee2874634b55ed45834054f5d9e

README.md
```diff
@@ -577,29 +577,15 @@ data "megaport_partner" "awshc" {
...
-  environment           = "production"
-  access_key            = "your-access-key"
-  secret_key            = "your-secret-key"
-  accept_purchase_terms = true
-  cancel_at_end_of_term = true  # Mark resources for end-of-term cancellation
-}
-```
-
-**Important notes:**
+**Billing impact:** Cancelling a Megaport resource before the end of its committed minimum term may incur early termination charges. Review your contract terms before destroying resources, especially in production.
```
mp1-tomclancy: How about "Review your contract terms before destroying resources in production."? 

The current statement seems to imply that users need to be worried about billing impacts in _any_ environment (i.e. staging).

mp1-tomclancy (changes requested):

README.md
Phil-Browne: Addressed in https://github.com/megaport/terraform-provider-megaport/pull/385/commits/2c4573242feecee76fea88aee6498ae4c52cb6fc

mp1-tomclancy (approved):

PR #390, opened 2026-05-26
ESD-1232: Document provider wait_time option

Phil-Browne:
Our docs didn't really explain the `wait_time` provider option, so users asking about it had nothing to go on. This adds proper coverage.

- New **Provisioning Wait Time** section in the provider index: what it does, the 10-minute default, when to raise it (MVEs, cloud VXCs, large applies), and what happens on timeout.
- Calls out the important gotcha: on timeout the order is already placed but isn't saved to state, so a re-apply creates a duplicate — recover by importing or cancelling.
- Adds a provider argument reference table and surfaces `wait_time` in the config example.
- Expands the `wait_time` schema description (notes it doesn't apply to IX, which uses a fixed 10-minute wait).

Docs regenerated via `go generate`; build, lint, and unit tests pass.


examples/provider/provider.tf
```diff
@@ -3,5 +3,5 @@ provider "megaport" {
   access_key            = "access_key"
   secret_key            = "secret_Key"
   accept_purchase_terms = true
-  wait_time             = 20
+  wait_time             = 20 # Minutes to wait for resources to provision (default 10)
```
megasergeyd, line 6: nitpick: "(default 20)"

megasergeyd (approved):

examples/provider/provider.tf
Phil-Browne: The provider default is actually 10, not 20 (`provider.go:163` sets `waitTime := 10` and the `wait_time` schema description reads `"Default value is 10"`); the example sets 20 only to show overriding it, so I'll keep `"(default 10)"`.

PR #395, opened 2026-06-11
ESD-1368 Mark all sensitive fields as such.

megasergeyd:
### User-Facing Summary

Mark all sensitive fields as such.
---

### Type of Change

- [ ] 🚀 New Feature
- [x] ✨ Enhancement
- [ ] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### Related Issues

---

### How to Test

---

### DIFF
--- a/docs/resources/mve.md
+++ b/docs/resources/mve.md
@@ -223,29 +223,29 @@ Required:
 
 Optional:
 
-- `account_key` (String) The account key for the vendor config. Enter the Account Key from Aruba Orchestrator. The key is linked to the Account Name. Required for Aruba MVE.
+- `account_key` (String, Sensitive) The account key for the vendor config. Enter the Account Key from Aruba Orchestrator. The key is linked to the Account Name. Required for Aruba MVE.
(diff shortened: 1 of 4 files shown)


Phil-Browne:
The MVE changes look correct, schema and regenerated docs are in sync (10 fields), and there's no state-compatibility risk since `Sensitive` only masks display.

Two things:

1. Title/description scope. This only touches MVE `vendor_config`, but "Mark all sensitive fields as such" reads as provider-wide. Worth narrowing the title/description to the MVE fields so it reflects what actually landed.

2. Follow-up ticket. While digging through the rest of the schema I noticed a few genuine secrets in the VXC partner configs that are still unmarked:
   - `vrouter_config` BGP `password` (`vxc_schemas.go:280`), this one's in the active config, so it's the one I'd most want covered
   - AWS `auth_key` (`vxc_schemas.go:42`)
   - Azure peer `shared_key` (`vxc_schemas.go:105`)

   I'd leave the deprecated `aEndPartnerConfigSchema` password alone since the v2 cleanup is rewriting that surface. Happy to raise a follow-up ticket for the three above so they don't get dropped.

Phil-Browne (approved):

PR #399, opened 2026-06-24
ESD-1539: Add IPsec tunnel configuration to the VXC resource

Phil-Browne:
Adds an `ip_sec_tunnel_options` block to each `vrouter_config` interface on `megaport_vxc`, so you can configure IPsec tunnels on an MCR vrouter interface. Bumps the megaportgo SDK to v1.14.0.

- New nested list `ip_sec_tunnel_options` (multiple tunnels per interface)
- `pre_shared_key` is sensitive; PSK and lifetimes are write-only (preserved from config, never drift)
- Plan-time validation of lifetime ranges and `phase2 < phase1`, folding in the API defaults when a lifetime is unset
- Unit tests for the mapping and validator, plus a `TF_ACC` acceptance test ordering MCR + IPsec add-on + VXC

Also marks the deprecated A-End BGP `password` field sensitive and drops two duplicate assignments spotted nearby.


Phil-Browne:
Reopened and reworked against the released megaportgo v1.14.1.

- Bumped megaportgo to v1.14.1.
- `ipsec_tunnel_options` is now a single nested object per interface (was a list). Multiple tunnels means multiple vRouter interfaces.
- Updated the schema, resource mapping, validator, tests, and docs to match.
- Fixed a `terraform destroy` race the acceptance test surfaced: deleting the IPsec add-on lowered its tunnel limit to 0 while the VXC's tunnel was still deprovisioning, returning a 400. The add-on Delete now retries until the tunnels clear (bounded by the existing 10-min wait).

Both adversarial review rounds plus a focused review of the teardown fix came back clean. Unit tests, lint, and the `TestAccMegaportVXC_IPsecTunnel` acceptance test pass against staging with clean teardown.

internal/provider/vxc_schemas.go
```diff
@@ -251,6 +251,54 @@ var (
...
+								},
+								"destination_ip_address": schema.StringAttribute{
+									Description: "Remote peer IPv4 address the tunnel connects to.",
+									Required:    true,
+								},
+								"pre_shared_key": schema.StringAttribute{
+									Description: "Pre-shared key used to authenticate the IPsec tunnel. Write-only: not returned by the API on read.",
+									Required:    true,
+									Sensitive:   true,
+								},
```
mp1-tomclancy, line 274: Can this be implemented as a [WriteOnly](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only) attribute? This would keep the PSK from being stored in state.

internal/provider/vxc_schemas.go
```diff
@@ -428,6 +476,7 @@ var (
 									},
 									"password": schema.StringAttribute{
 										Description: "The password of the BGP connection.",
+										Sensitive:   true,
 										Optional:    true,
 									},
```
mp1-tomclancy, line 482: Same as `pre_shared_key`, could this also be `WriteOnly`?

internal/provider/vxc_schemas.go
Phil-Browne: Done in 865afc7. `pre_shared_key` is now a Terraform 1.11+ write-only argument (`Sensitive` + `WriteOnly`), so it is never written to the plan or state. It is read from the configuration when the tunnel is ordered and sourced separately from the plan-derived model, so it stays out of the partner-config object that becomes state. Added a `TestCheckNoResourceAttr` assertion that it is absent from state, and the docs now render the `[Write-only]` marker.

internal/provider/vxc_schemas.go
Phil-Browne: Left this one as `Sensitive` rather than write-only. Unlike `pre_shared_key` (new in this PR), the BGP `password` is an already-released attribute, so converting it to write-only is a breaking change: it forces consumers onto Terraform 1.11+/OpenTofu 1.10+ and drops the value out of existing state on upgrade. This field is also on the deprecated `a_end_partner_config` schema. Happy to do it (along with the equivalent `vrouter_config` BGP password) as a separate, clearly flagged change if we want to accept that breaking change, but it felt out of scope here.

mp1-tomclancy: Good point re. deprecation, that seems like the right call.

mp1-tomclancy (approved):

PR #410, opened 2026-07-08
chore: deprecate legacy computed metadata on port and lag_port

Phil-Browne:
Marks the read-only bookkeeping attributes on `megaport_port` and `megaport_lag_port` as deprecated. These echo API metadata and clutter state without being useful in config.

Deprecated (16 per resource): `product_id`, `provisioning_status`, `create_date`, `created_by`, `last_updated`, `live_date`, `terminate_date`, `contract_start_date`, `contract_end_date`, `market`, `usage_algorithm`, `virtual`, `locked`, `cancelable`, `vxc_permitted`, `vxc_auto_approval`.

Non-breaking: the attributes still populate as before; anyone referencing them now gets a deprecation warning ahead of removal in a future major release. Docs regenerated. This is the first of a per-resource series (mcr/mve/vxc to follow).

### DIFF
--- a/docs/resources/lag_port.md
+++ b/docs/resources/lag_port.md
@@ -45,26 +45,26 @@ resource "megaport_lag_port" "lag_port" {
 
 ### Read-Only
 
-- `cancelable` (Boolean) Whether the product is cancelable.
+- `cancelable` (Boolean, Deprecated) Whether the product is cancelable.
 - `company_uid` (String) The unique identifier of the company.
-- `contract_end_date` (String) The date the contract ends.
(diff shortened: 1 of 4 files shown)


Phil-Browne:
Closing this. v2 is shipping once with a clean final schema, so there's no intermediate release where a deprecation warning would ever be seen. Marking these fields deprecated and then deleting them in the same unreleased branch doesn't warn any real user of anything, it's just extra review overhead and a source of merge conflicts with #419. The removal itself is already covered end to end in #419.

PR #411, opened 2026-07-08
chore: deprecate legacy computed metadata on mve

Phil-Browne:
Adds a `DeprecationMessage` to 21 read-only, computed metadata attributes on the `megaport_mve` resource: `last_updated`, `product_id`, `product_type`, `provisioning_status`, `create_date`, `created_by`, `terminate_date`, `live_date`, `contract_start_date`, `contract_end_date`, `market`, `usage_algorithm`, `secondary_name`, `company_name`, `virtual`, `locked`, `admin_locked`, `cancelable`, `buyout_port`, `vxc_permitted`, and `vxc_auto_approval`.

This is non-breaking: the attributes still populate and behave exactly as before, users just get a deprecation warning ahead of removal in a future major release. Docs regenerated. This is part of a per-resource series applying the same deprecations across resources.

### DIFF
--- a/docs/resources/mve.md
+++ b/docs/resources/mve.md
@@ -184,33 +184,33 @@ resource "megaport_mve" "mve_sixwind_dynamic" {
 
 ### Read-Only
 
-- `admin_locked` (Boolean) Whether the MVE is admin locked.
+- `admin_locked` (Boolean, Deprecated) Whether the MVE is admin locked.
 - `attribute_tags` (Map of String) The attribute tags of the MVE.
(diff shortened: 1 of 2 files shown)


Phil-Browne:
Closing this. v2 is shipping once with a clean final schema, so there's no intermediate release where a deprecation warning would ever be seen. Marking these fields deprecated and then deleting them in the same unreleased branch doesn't warn any real user of anything, it's just extra review overhead and a source of merge conflicts with #422. The removal itself is already covered end to end in #422.

PR #413, opened 2026-07-08
chore: deprecate legacy computed metadata on vxc

Phil-Browne:
Adds a deprecation warning to a set of read-only computed metadata attributes on the `megaport_vxc` resource. This is non-breaking: the attributes still populate exactly as before and only emit a deprecation warning plus a doc marker, ahead of removal in a future major release. This is part of a per-resource series doing the same cleanup across the provider.

Deprecated fields: `last_updated`, `product_id`, `product_type`, `provisioning_status`, `create_date`, `created_by`, `live_date`, `contract_start_date`, `contract_end_date`, `usage_algorithm`, `secondary_name`, `company_name`, `locked`, `admin_locked`, `cancelable`.

Docs regenerated via `tfplugindocs`.

### DIFF
--- a/docs/resources/vxc.md
+++ b/docs/resources/vxc.md
@@ -307,27 +307,27 @@ resource "megaport_vxc" "service_key_vxc" {
 
 ### Read-Only
 
-- `admin_locked` (Boolean) Whether the product is admin locked.
+- `admin_locked` (Boolean, Deprecated) Whether the product is admin locked.
 - `attribute_tags` (Map of String) The attribute tags associated with the product.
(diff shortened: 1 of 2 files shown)


Phil-Browne:
Closing this. v2 is shipping once with a clean final schema, so there's no intermediate release where a deprecation warning would ever be seen. Marking these fields deprecated and then deleting them in the same unreleased branch doesn't warn any real user of anything, it's just extra review overhead and a source of merge conflicts with #420. The removal itself is already covered end to end in #420.

PR #350, opened 2026-04-14
ESD-791(v2): add GNUmakefile with standard provider targets

Phil-Browne:
Adds a `GNUmakefile` following the HashiCorp terraform-provider-scaffolding convention. This provides standard targets that every Terraform provider contributor and CI system expects.

## Targets

| Target | Description |
|---|---|
| `make build` | Build the provider |
| `make install` | Install to `$GOPATH/bin` |
| `make test` | Run unit tests with coverage |
| `make testacc` | Run acceptance tests (`TF_ACC=1`) |
| `make lint` | Run golangci-lint |
| `make generate` | Regenerate docs via `go generate` |
| `make fmt` | Format Go and Terraform files |
| `make clean` | Clean build artifacts |

## Test plan

- [x] `make build` succeeds
- [x] `make fmt` runs without error

PR #415, opened 2026-07-08
chore(v2): add GNUmakefile with standard provider targets

Phil-Browne:
Adds a `GNUmakefile` with the standard provider targets: `build`, `install`, `test`, `testacc`, `lint`, `generate`, `fmt`, `clean`. Gives common workflows a single entry point.

Re-cut of the stalled `v2/pr-21-makefile` PR (#350) against `main`, dropping the v2 module-path and dependency churn it carried.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,0 +1,35 @@
+default: build
+
+build:
+	go build -v ./...
+
+install:
+	go install -v ./...
+
+test:
+	go test -v -timeout=30m -cover ./internal/provider/
+
+testacc:
+ifndef MEGAPORT_ACCESS_KEY
+	$(error MEGAPORT_ACCESS_KEY is not set)
+endif
+ifndef MEGAPORT_SECRET_KEY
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -0,0 +1,35 @@
...
+
+lint:
+	golangci-lint run --timeout=10m
+
+generate:
+	go generate ./...
+
+fmt:
+	gofmt -w .
+	@command -v terraform >/dev/null 2>&1 && terraform fmt -recursive examples/ || echo "Skipping terraform fmt: terraform CLI not found"
```
mp1-tomclancy: NIT: message could be misleading if `terraform fmt` fails for any reason other than missing tf cli

mp1-tomclancy (approved):

GNUmakefile
Phil-Browne: Fixed in https://github.com/megaport/terraform-provider-megaport/pull/415#:~:text=Verified-,e632cb2,-Merge%20info

PR #336, opened 2026-04-05
ESD-777(v2): remove noise fields from port and LAG port resources

Phil-Browne:
The `megaport_port` and `megaport_lag_port` resources exposed a large number of computed fields that are read-only API metadata with no user value. These fields cluttered `terraform plan` output and created confusion about what is actually managed vs observed. This PR establishes the pattern for all subsequent V2 resource cleanup.

## Removed fields (both resources)

`product_id`, `provisioning_status`, `usage_algorithm`, `vxc_permitted`, `vxc_auto_approval`, `virtual`, `locked`, `cancelable`, `live_date`, `create_date`, `contract_start_date`, `contract_end_date`, `created_by`, `terminate_date`, `market`

Each field was removed from the schema, model struct, and `fromAPIPort` mapping function.

## Retained fields

`product_uid`, `product_name`, `port_speed`, `location_id`, `contract_term_months`, `marketplace_visibility`, `diversity_zone`, `cost_centre`, `promo_code`, `company_uid`, `resources` (interface demarcation/up), `resource_tags`, plus LAG-specific `lag_count` and `lag_port_uids`.

## Verification

- `go build -v ./...` passes
- `go test -v -timeout=30m ./internal/provider/` passes

PR #419, opened 2026-07-08
ESD-777-feat(v2): remove deprecated metadata fields from port and LAG port resources

Phil-Browne:
Removes the 16 computed metadata fields deprecated in #410 (last_updated, product_id, provisioning_status, create_date, created_by, terminate_date, live_date, market, usage_algorithm, contract_start_date, contract_end_date, vxc_permitted, vxc_auto_approval, virtual, locked, cancelable) from both `megaport_port` and `megaport_lag_port`, and adds V1-to-V2 state migration. Re-cut of the retired #336.

- Drop the attributes from the schema, model, and fromAPI/Create/Update paths of both resources
- Add a `ResourceWithMoveState` implementation (`port_move_state.go`) that builds a V2 model from raw V1 state, silently dropping removed fields, with unit tests
- Trim acceptance test checks and `ImportStateVerifyIgnore` lists accordingly
- Regenerate `docs/resources/port.md` and `docs/resources/lag_port.md`

This will conflict with #410 by design since the removal deletes the deprecated attribute blocks; whichever merges second rebases.


Phil-Browne:
Per the review feedback on #421: Terraform core only dispatches `MoveResourceState` to the provider when a `moved` block crosses resource types or providers. `prepareCrossTypeMove` (internal/refactoring/cross_provider_move.go) returns early for a same-provider, same-type move, so the `megaport_port` and `megaport_lag_port` StateMovers here (which guarded on exactly that same-type case) were dead code and could never be invoked.

Removed the movers and the `resource.ResourceWithMoveState` assertions. The V1 to V2 upgrade needs no provider code at all: the framework's `UpgradeResourceState` same-version passthrough decodes prior state against the current schema with `IgnoreUndefinedAttributes: true`, silently dropping the removed attributes on the first plan/refresh with the v2 provider. Replaced the mover tests with `port_state_upgrade_test.go`, which replicates that passthrough and proves full V1 states (removed fields populated) still decode against the V2 schema.

megasergeyd (approved):

megasergeyd (approved):

PR #339, opened 2026-04-08
ESD-779(v2): VXC resource V2 comprehensive cleanup

Phil-Browne:
This is a comprehensive V2 cleanup of the `megaport_vxc` resource — the largest resource in the provider. All changes are breaking and intentional for the V2 major version.

## What changed

**Phase 1 — Noise field removal**
Removed Computed-only fields that add no value for practitioners. Top-level: `product_id`, `product_type`, `provisioning_status`, `secondary_name`, `usage_algorithm`, `company_name`, `locked`, `admin_locked`, `cancelable`, `live_date`, `create_date`, `contract_start_date`, `contract_end_date`. End configs: `owner_uid`, `product_name`, `secondary_name`, `location`, `location_id`.

**Phase 2 — VLAN consolidation**
Dropped `ordered_vlan`. `vlan` is now `Optional+Computed`, validated `Between(-1, 4093)` and `NoneOf(1)`: 0 requests auto-assignment, -1 is untagged, 2–4093 is a specific VLAN.

**Phase 3 — End UID rename**
- `requested_product_uid` → `product_uid` (user input)
- `current_product_uid` → `assigned_product_uid` (API output — may differ for partner ports)

**Phase 4 — Partner config restructure**
- `a_end` / `b_end` → `a_end_config` / `b_end_config`
- Partner configs (`aws_config`, `azure_config`, `google_config`, `oracle_config`, `ibm_config`, `vrouter_config`) now nest directly inside the relevant end config; the top-level `a_end_partner_config` / `b_end_partner_config` discriminator blocks are gone
- Azure: dropped `port_choice` — supply the port UID directly via `product_uid`
- Transit: `partner = "transit"` becomes `transit = true` on `b_end_config`
- Separate A-end / B-end model structs reflect the API rule that CSP partners only connect on the B-end

**Phase 5 — WriteOnly secrets + CSP data source**
- `aws_config.auth_key`, `azure_config.service_key`, and `vrouter_config…bgp_connections[].password` are now `WriteOnly` — never persisted to state
- `csp_connections` moved off the resource into a new `megaport_vxc_csp_connection` data source (takes a `vxc_uid`, returns connection details)

## Bug and security fixes

- **Spurious destroy+recreate of AWS/Azure VXCs**: WriteOnly `auth_key`/`service_key` were triggering RequiresReplace on every apply; now nullified before comparison.
- **Spurious vrouter updates**: BGP passwords are stripped before plan/state comparison, so unchanged configs no longer make API calls.
[...]

PR #420, opened 2026-07-08
ESD-779-feat(v2): remove deprecated metadata fields from VXC resource

Phil-Browne:
> Split: the MCR inline `prefix_filter_lists` state-upgrade change that previously shared this branch now lives in #433. This PR is VXC-only.

Removes the 15 legacy computed metadata attributes that were deprecated in #413 from the `megaport_vxc` resource. This replaces the removal slice of the abandoned #339 (which was built on a dead v2 stack).

Changes:
- Drop these attributes from the schema, model struct, and API mapping: `last_updated`, `product_id`, `product_type`, `provisioning_status`, `secondary_name`, `usage_algorithm`, `created_by`, `live_date`, `create_date`, `contract_start_date`, `contract_end_date`, `company_name`, `locked`, `admin_locked`, `cancelable`.
- No state migrator is needed. Because the change only removes attributes, the framework's built-in `UpgradeResourceState` passthrough decodes existing V1 state against the V2 schema (with `IgnoreUndefinedAttributes`), dropping the removed attributes and carrying everything else over unchanged. New tests in `vxc_state_upgrade_test.go` prove real V1 state, including the nested a_end/b_end/partner objects, still decodes cleanly against the V2 schema.
- Regenerate `docs/resources/vxc.md`.

Out of scope and untouched: partner config schemas, the `requested_product_uid`/`current_product_uid` rotation logic, and the nested a_end/b_end `secondary_name` (a different field from the removed top-level one). The API-object provisioning-status checks in Read and the provisioning wait are preserved.


Phil-Browne:
Per the review feedback on #421: Terraform core only dispatches `MoveResourceState` to the provider when a `moved` block crosses resource types or providers (`prepareCrossTypeMove` in `internal/refactoring/cross_provider_move.go` returns early for a same-provider, same-type move). The `megaport_vxc` StateMover guarded on exactly that same-type case, so it was dead code and is removed.

The V1 to V2 upgrade needs no provider code at all: the framework's `UpgradeResourceState` same-version passthrough decodes prior state against the current schema with `IgnoreUndefinedAttributes: true`, dropping the removed attributes on the first plan/refresh after upgrading the provider. Replaced the mover tests with `vxc_state_upgrade_test.go`, which replicates that passthrough over a full V1 state (removed fields populated, plus the nested `a_end`/`b_end` and partner config objects) and proves no kept attribute changed type.

PR #412, opened 2026-07-08
chore: deprecate legacy computed metadata on mcr

Phil-Browne:
Adds a `DeprecationMessage` to the read-only bookkeeping attributes on the `megaport_mcr` resource. This is non-breaking: the fields still populate and keep working, so this only surfaces a deprecation warning ahead of removal in a future major release. This is part of a per-resource series doing the same cleanup across the provider.

Deprecated fields (24): `last_updated`, `product_id`, `product_type`, `provisioning_status`, `create_date`, `created_by`, `terminate_date`, `live_date`, `contract_start_date`, `contract_end_date`, `market`, `usage_algorithm`, `secondary_name`, `company_name`, `virtual`, `locked`, `admin_locked`, `cancelable`, `buyout_port`, `vxc_permitted`, `vxc_auto_approval`, `aggregation_id`, `lag_id`, `lag_primary`.

Docs regenerated (`docs/resources/mcr.md`).

PR #337, opened 2026-04-05
feat: remove noise fields and inline prefix filter lists from MCR resource

Phil-Browne:
Cleans up the \`megaport_mcr\` resource for V2 by removing fields that don't provide value to Terraform users, removing the deprecated inline prefix filter list management, and adding IPSec add-on support as a new standalone resource.

**Noise fields removed (22):** \`product_id\`, \`product_type\`, \`provisioning_status\`, \`create_date\`, \`created_by\`, \`terminate_date\`, \`live_date\`, \`market\`, \`usage_algorithm\`, \`company_name\`, \`contract_start_date\`, \`contract_end_date\`, \`secondary_name\`, \`lag_primary\`, \`lag_id\`, \`aggregation_id\`, \`vxc_permitted\`, \`vxc_auto_approval\`, \`virtual\`, \`buyout_port\`, \`locked\`, \`admin_locked\`, \`cancelable\`.

**Inline prefix_filter_lists removed:** The deprecated inline \`prefix_filter_lists\` attribute has been removed entirely. Customers should use the standalone \`megaport_mcr_prefix_filter_list\` resource instead. This removes the inline CRUD logic, validation/deprecation warnings, the \`RateLimiter\`, and \`mcr_resource_utils.go\`.

**Fields retained:** \`product_uid\`, \`product_name\`, \`port_speed\`, \`location_id\`, \`contract_term_months\`, \`cost_centre\`, \`diversity_zone\`, \`promo_code\`, \`asn\`, \`company_uid\`, \`marketplace_visibility\`, \`attribute_tags\`, \`resource_tags\`.

**New resource: \`megaport_mcr_ipsec_addon\`** (from megaportgo PR #113): Implements IPSec tunnel add-on support as a standalone resource, following the same pattern as \`megaport_mcr_prefix_filter_list\`. Embedding add-on state in the MCR resource caused force-replace on every tunnel count change, so a separate resource is the correct design.

- \`mcr_id\` — references the parent MCR UID (RequiresReplace)
- \`tunnel_count\` — required, valid values: 10, 20, 30
- \`add_on_uid\` — computed, assigned by API
- Create → \`UpdateMCRWithAddOn\`; Update → \`UpdateMCRIPsecAddOn\`; Delete → set tunnel count to 0 (disable)
- Polls until MCR reaches ready state after each mutation (add-on API has no built-in wait)
- Import via composite \`mcr_uid:add_on_uid\` format
[...]

PR #421, opened 2026-07-08
feat(v2): remove deprecated legacy metadata fields from megaport_mcr

Phil-Browne:
Removes the deprecated read-only metadata fields and the inline `prefix_filter_lists` attribute from the `megaport_mcr` resource. All of these were marked deprecated in #412. Replaces #337, which was built on the now-dead v2 stack.

Removed attributes (schema, model field, and all create/read/update wiring):

`last_updated`, `product_id`, `product_type`, `provisioning_status`, `create_date`, `created_by`, `terminate_date`, `live_date`, `market`, `usage_algorithm`, `contract_start_date`, `contract_end_date`, `secondary_name`, `lag_primary`, `lag_id`, `aggregation_id`, `company_name`, `vxc_permitted`, `vxc_auto_approval`, `virtual`, `buyout_port`, `locked`, `admin_locked`, `cancelable`, and the inline `prefix_filter_lists` (use the standalone `megaport_mcr_prefix_filter_list` resource instead).

Other changes:

- Add a V1-to-V2 state mover (`mcr_move_state.go`) so existing state migrates cleanly, dropping the removed fields.
- Delete `mcr_resource_utils.go` (only supported the inline prefix filter list plan modifier).
- Regenerate `docs/resources/mcr.md`.


internal/provider/mcr_move_state.go
```diff
@@ -0,0 +1,41 @@
+package provider
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func (r *mcrResource) MoveState(ctx context.Context) []resource.StateMover {
```
mp1-tomclancy: It's not really clear to me how this method comes into play. My reading of the [docs](https://developer.hashicorp.com/terraform/plugin/framework/resources/state-move) is that this method will be invoked when:
- a `moved` block is explicitly defined  by the user
- the source and target resource types (or providers) do not match

Since the second condition isn't true here (same provider/same resource name), I don't think this will get called even if the user creates a `moved` block.

Phil-Browne: Confirmed against Terraform core: `prepareCrossTypeMove` (internal/refactoring/cross_provider_move.go) returns early when the source and target share the same provider and resource type, so `MoveResourceState` would never be dispatched for this mover. It was dead code.

The V1 to V2 upgrade also turns out to need no provider code at all: the framework's `UpgradeResourceState` passthrough decodes prior state against the current schema with `IgnoreUndefinedAttributes: true` (fwserver/server_upgraderesourcestate.go), so the removed attributes are silently dropped on the first plan/refresh with the new provider.

Removed `mcr_move_state.go` and replaced the tests with `mcr_state_upgrade_test.go`, which replicates that passthrough and proves a full V1 state (removed fields populated, including `prefix_filter_lists`) still decodes against the V2 schema. Note the same MoveState pattern exists in #419, #420, and #422 and will get the same fix.

mp1-tomclancy: > The V1 to V2 upgrade also turns out to need no provider code at all: the framework's UpgradeResourceState passthrough decodes prior state against the current schema with IgnoreUndefinedAttributes: true (fwserver/server_upgraderesourcestate.go), so the removed attributes are silently dropped on the first plan/refresh with the new provider.

Let me know if I've got this wrong, but I'm inferring the migration plan looks like:
- User updates to `terraform-provider-megaport` v2
- The user manually deletes the removed fields from their terraform config
- The provider automatically removes the old fields from state (via the `UpgradeResourceState` passthrough)

I'm guessing `prefix_filter_lists`, would need to be migrated to standalone resources prior to the V2 upgrade?

Phil-Browne: Thanks for this, good sanity checks for sure. Dug a bit deeper and had Fable do a pretty thorough check as well. Short version is you are mostly right, two corrections:

- Steps 1 and 3: correct. The state cleanup is automatic on the first plan/refresh with v2, no state surgery needed.
- Step 2 barely applies: everything removed except prefix_filter_lists was computed-only, so none of it can appear in config as an argument. What users actually clean up is *references* to the removed attributes (outputs/expressions like `megaport_mcr.x.provisioning_status`), which fail plan validation with an "Unsupported attribute" error.
- `prefix_filter_lists` is the other way around: migrating before the upgrade isn't required, and after is safer. The upgrade just drops the attribute from schema and state; the API-side lists survive untouched (v2's MCR code has no prefix-filter-list handling left, so nothing can delete them). Users then import each list into a standalone `megaport_mcr_prefix_filter_list` with `terraform import ... "mcr_uid:prefix_list_id"`, which keeps the list IDs that BGP connection filters reference. Migrating on v1 first has a footgun: an explicit `prefix_filter_lists = []` deletes every list on the MCR via the API.

I've put this in a draft upgrade guide: #427 (draft until the removal PRs merge).

mp1-tomclancy: Sounds good, thanks for the clarifiation. 

r.e. the `prefix_filter lists`, I'll leave a quick comment on #427.

mp1-tomclancy (approved):

PR #422, opened 2026-07-08
feat(v2): MVE per-vendor config blocks and legacy field removal

Phil-Browne:
Re-cut of #352 (that PR's stack is dead). Two changes to `megaport_mve`.

**Per-vendor config split.** The single union `vendor_config` block is replaced by one Optional block per vendor: `aruba_config`, `aviatrix_config`, `cisco_config`, `fortinet_config`, `meraki_config`, `palo_alto_config`, `prisma_config`, `sixwind_config`, `versa_config`, `vmware_config`. `ExactlyOneOf` enforces that one and only one is set, and the vendor is inferred from the block, so the old `vendor` selector string is gone. Only `image_id` and `product_size` are required per block; the API stays the source of truth for the rest.

**Legacy field removal.** Drops the 21 deprecated read-only fields (`last_updated`, `product_id`, `provisioning_status`, `create_date`, `created_by`, `terminate_date`, `live_date`, `market`, `product_type`, `usage_algorithm`, `contract_start_date`, `contract_end_date`, `vxc_permitted`, `vxc_auto_approval`, `secondary_name`, `company_name`, `virtual`, `buyout_port`, `locked`, `admin_locked`, `cancelable`) plus `vnics.vlan`.

- `admin_password` (Cisco FTDv, write-only + sensitive) is preserved and re-read from config in Create.
- Adds a `ResourceWithMoveState` migrator for V1 to V2: the V2 schema doubles as the source schema, so the framework decode drops the removed fields, the old union block, and `vnics.vlan`; per-vendor blocks reconcile from config on the next apply.
- Sensitive flags unchanged from the current schema; no fields promoted to Sensitive.

Verified: `go build`, `golangci-lint` (0 issues), unit tests, `gofmt`, and `go generate` (only `docs/resources/mve.md` changes) all clean.

### DIFF
--- a/docs/resources/mve.md
+++ b/docs/resources/mve.md
@@ -172,81 +172,193 @@ resource "megaport_mve" "mve_sixwind_dynamic" {
 - `contract_term_months` (Number) The term of the contract in months: valid values are 1, 12, 24, 36, 48, and 60. To set the product to a month-to-month contract with no minimum term, set the value to 1.
 - `location_id` (Number) The numeric location ID of the product. This value can be retrieved from the data source megaport_location.
 - `product_name` (String) The name of the MVE.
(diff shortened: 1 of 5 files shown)


Phil-Browne:
Splitting this into two stacked PRs for reviewability, matching how the other three resources handled removal as its own PR (#419 / #420 / #421):

- Legacy field removal: https://github.com/megaport/terraform-provider-megaport/pull/423
- Vendor-config split (stacked on the removal PR): https://github.com/megaport/terraform-provider-megaport/pull/424

Closing in favor of those two. The `feat/mve-v2-vendor-split` branch is kept.

PR #423, opened 2026-07-08
feat(v2): remove deprecated metadata fields from MVE resource

Phil-Browne:
> **Merge order: this PR is #1 of 2. Merge this before #424.** The vendor-config split (#424) is stacked on this branch and targets it, so #424 cannot merge until this lands. After this merges, retarget #424 to `feat/v2`.

Sibling to #419 / #420 / #421: removes the deprecated read-only metadata fields from `megaport_mve`, matching how the other resources handled removal on their own.

- Drops 21 deprecated fields: `last_updated`, `product_id`, `provisioning_status`, `create_date`, `created_by`, `terminate_date`, `live_date`, `market`, `product_type`, `usage_algorithm`, `contract_start_date`, `contract_end_date`, `vxc_permitted`, `vxc_auto_approval`, `secondary_name`, `company_name`, `virtual`, `buyout_port`, `locked`, `admin_locked`, `cancelable`.
- Drops the `vnics.vlan` nested attribute (vNIC blocks are now just `description`).
- Adds a V1-to-V2 state migrator: removed fields and `vnics.vlan` drop on decode; everything else carries over unchanged.
- Keeps the Read decommissioned-status guard.

`vendor_config` is untouched here (stays a single union block). The vendor-config split into per-vendor blocks is the follow-up PR #424, stacked on this one.

Verified: `go build`, docs regen (only `docs/resources/mve.md`, removed fields gone, `vendor_config` still a union), `golangci-lint` (0 issues), unit tests, and `gofmt` all clean.


Phil-Browne:
Per review feedback on #421: Terraform core only dispatches `MoveResourceState` when a moved block crosses resource types or providers (`prepareCrossTypeMove` in internal/refactoring/cross_provider_move.go returns early for same-provider, same-type moves), so the same-type MoveState migrator here was dead code and is removed.

The V1 to V2 upgrade needs no provider code: the framework's `UpgradeResourceState` passthrough decodes prior state against the current schema with `IgnoreUndefinedAttributes: true`, dropping the removed attributes (including the nested `vnics.vlan`). Replaced the mover tests with `mve_state_upgrade_test.go`, which exercises that passthrough against a full V1 state.

ianb-mp (approved):

PR #424, opened 2026-07-08
ESD-793-feat(v2): split MVE vendor_config into per-vendor blocks (BREAKING)

Phil-Browne:
> **Merge order: this PR is #2 of 2. Merge #423 first, then this.** This branch is stacked on #423 and currently targets `feat/remove-mve-legacy-fields`, so the diff below shows only the incremental vendor-split change. Once #423 merges, retarget this PR to `feat/v2` (GitHub usually does this automatically when the base branch is deleted, but confirm).

This is the vendor-config-split half of the old #422, split out for reviewability.

Swaps the single union `vendor_config` block for 10 Optional per-vendor blocks: `aruba_config`, `aviatrix_config`, `cisco_config`, `fortinet_config`, `meraki_config`, `palo_alto_config`, `prisma_config`, `sixwind_config`, `versa_config`, `vmware_config`.

- `objectvalidator.ExactlyOneOf` enforces exactly one block; the vendor is inferred from the set block, so the old `vendor` selector string is gone.
- Only `image_id` and `product_size` are required per block; the API stays the source of truth for everything else.
- `admin_password` (Cisco FTDv, write-only + sensitive) is preserved and re-read from config in Create.
- Per-block `UseStateForUnknown` plus case-insensitive (`strings.EqualFold`) vendor/size compares in ModifyPlan, no object-level RequiresReplace.
- Extends the V1-to-V2 migrator: the old union block drops on decode; per-vendor blocks reconcile from config on the next apply.

BREAKING: existing `vendor_config` blocks must be rewritten as the matching per-vendor block.

Verified: `go build`, docs regen (per-vendor blocks and `admin_password` shown write-only + sensitive), `golangci-lint` (0 issues), unit tests, and `gofmt` all clean.


internal/provider/mve_resource.go
```diff
@@ -64,41 +76,460 @@ type mveNetworkInterfaceModel struct {
...
+	dst.PaloAltoConfig = src.PaloAltoConfig
+	dst.PrismaConfig = src.PrismaConfig
+	dst.SixwindConfig = src.SixwindConfig
+	dst.VersaConfig = src.VersaConfig
+	dst.VmwareConfig = src.VmwareConfig
+}
+
+// vendorNameFromModel returns the canonical vendor name based on which config
+// block is set in the model, or "" if none is set.
+func vendorNameFromModel(m mveResourceModel) string {
```
projectsuperba, line 460: There's quite a few places in this file where all vendors are cased out individually which is high maintenance if one is removed/added in the future. I'd suggest a central registry for configuring these vendors' fields and then functions such as `vendorNameFromModel()` can be collapsed to generic loops like so:

```golang
func vendorNameFromModel(m mveResourceModel) string {
	for _, v := range mveVendors {
		if objectSet(v.get(&m)) {
			return v.name
		}
	}
	return ""
}
```

and possible corresponding registry setup:

```golang
type mveVendorSpec struct {
	name     string                       // canonical name, e.g. "cisco", "6wind"
	attrName string                       // schema/tfsdk name, e.g. "cisco_config"
	get      func(*mveResourceModel) types.Object
	set      func(*mveResourceModel, types.Object)
	toAPI    func(context.Context, types.Object, string) (megaport.VendorConfig, diag.Diagnostics)
}

// Single source of truth. Adding a vendor = one entry here.
var mveVendors = []mveVendorSpec{
	{
		name: "aruba", attrName: "aruba_config",
		get: func(m *mveResourceModel) types.Object { return m.ArubaConfig },
		set: func(m *mveResourceModel, o types.Object) { m.ArubaConfig = o },
		toAPI: func(ctx context.Context, o types.Object, _ string) (megaport.VendorConfig, diag.Diagnostics) {
			var cfg arubaConfigModel
			diags := o.As(ctx, &cfg, basetypes.ObjectAsOptions{})
			if diags.HasError() {
				return nil, diags
			}
			return &megaport.ArubaConfig{
				Vendor: "aruba", ImageID: int(cfg.ImageID.ValueInt64()),
				ProductSize: cfg.ProductSize.ValueString(), MVELabel: cfg.MVELabel.ValueString(),
				AccountName: cfg.AccountName.ValueString(), AccountKey: cfg.AccountKey.ValueString(),
				SystemTag: cfg.SystemTag.ValueString(),
			}, diags
		},
	},
	// aviatrix, cisco (uses the adminPassword arg), fortinet, … one entry each.
}
```

another function example, passing `adminPassword` through for Cisco:

```golang
func toAPIVendorConfigFromModel(ctx context.Context, m *mveResourceModel, adminPassword string) (megaport.VendorConfig, diag.Diagnostics) {
	name := vendorNameFromModel(*m)
	spec, ok := specForVendor(name)
	if !ok {
		var d diag.Diagnostics
		d.AddError("Invalid vendor configuration", "Exactly one vendor configuration block must be set.")
		return nil, d
	}
	return spec.toAPI(ctx, spec.get(m), adminPassword)
}

```

Phil-Browne: Good call. Pushed 2c933b1: collapsed the vendor helpers (vendorObjects, copyVendorConfigs, vendorNameFromModel, vendorConfigPath, blockForVendor, toAPIVendorConfigFromModel) into loops over a single `mveVendors` registry, following the shape you suggested. Adding or removing a vendor is now a one-entry edit to that slice.

Phil-Browne:
Merged the updated #423 base, which carries the dead-MoveState fix. Per the review feedback on #421: Terraform core only dispatches `MoveResourceState` when a `moved` block crosses resource types or providers (`prepareCrossTypeMove` in `internal/refactoring/cross_provider_move.go` returns early for a same-provider, same-type move), so the `megaport_mve` StateMover was dead code. Removed the mover, its file, and the `ResourceWithMoveState` interface assertion added by this branch's vendor split.

The V1 to V2 upgrade needs no provider code: the framework's `UpgradeResourceState` passthrough decodes prior state against the current schema with `IgnoreUndefinedAttributes: true`. `mve_state_upgrade_test.go` exercises that passthrough over a full V1 state and confirms the old union `vendor_config` drops, the per-vendor config blocks decode to null, and the removed nested `vnics.vlan` attribute is dropped, all with no kept attribute changing type.

projectsuperba (approved):

PR #425, opened 2026-07-09
ESD-1620: normalize inner_vlan -1/0 in verifyUpdateApplied

Phil-Browne:
## Summary
Update requests that set `inner_vlan = -1` (untagged) return `0` from the API for that end. `verifyUpdateApplied` compared these raw values, so it never matched, and `Update()` would spin for the full 120s timeout logging a spurious "VXC Update Propagation Delay" warning even though the update applied correctly. Mirrors the normalization `fromAPIVXC` already does on Read.

- Add `innerVLANMatches` helper: requested `-1` matches API-returned `0`, otherwise plain equality.
- Use it in `verifyUpdateApplied` for `AEndInnerVLAN`/`BEndInnerVLAN` (outer VLAN untouched, since `0` there means auto-assign, not untagged).
- Fix the same raw comparison in `Update()`'s A-End/B-End Inner VLAN Mismatch warnings, which had the identical bug and could fire a false mismatch whenever the update wait timed out for an unrelated reason.
- Add table-driven unit tests for `verifyUpdateApplied` (none existed before).
- Remove the "known gap" comment on `TestAccMegaportMVE_to_MVE_VXC` documenting this exact bug.

## Test plan
- [x] `go build -v ./...`
- [x] `go test -v -timeout=30m -run TestVerifyUpdateApplied ./internal/provider/`
- [x] `go test -timeout=30m ./internal/provider/`
- [x] `golangci-lint run --timeout=10m`
- [x] `TestAccMegaportMVE_to_MVE_VXC` against staging: passes in ~4.2 min, no propagation-delay or VLAN-mismatch warning

### DIFF
--- a/internal/provider/vxc_resource_utils_test.go
+++ b/internal/provider/vxc_resource_utils_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	megaport "github.com/megaport/megaportgo"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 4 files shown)


internal/provider/vxc_resource_utils_test.go
```diff
@@ -158,3 +159,101 @@ func TestIPSecPhaseLifetimeValidator(t *testing.T) {
...
+
+	r := &vxcResource{}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.want, r.verifyUpdateApplied(tc.vxc, tc.updateReq))
+		})
+	}
+}
+
+func strPtr(s string) *string { return &s }
```
projectsuperba: Use `megaport.PtrTo()` for this

internal/provider/vxc_resource_utils_test.go
```diff
@@ -158,3 +159,101 @@ func TestIPSecPhaseLifetimeValidator(t *testing.T) {
 		})
 	}
 }
+
+// TestVerifyUpdateApplied covers the -1 (requested untagged) / 0 (API
+// returned untagged) inner VLAN normalization, a genuine mismatch that must
+// still fail, and unrelated fields being unaffected by the normalization.
+func TestVerifyUpdateApplied(t *testing.T) {
+	intPtr := func(v int) *int { return &v }
```
projectsuperba: `megaport.PtrTo()`

projectsuperba (changes requested):

internal/provider/vxc_resource_utils_test.go
Phil-Browne: Done, pushed in bf99e1a.

internal/provider/vxc_resource_utils_test.go
Phil-Browne: Done, pushed in bf99e1a.

projectsuperba (approved):

PR #427, opened 2026-07-10
docs: add v2 upgrade guide

Phil-Browne:
Adds a registry guide (`templates/guides/v2_upgrade.md.tmpl`) walking users through the v1 to v2 upgrade:

- Removed read-only attributes per resource, and that the fix is removing references (state cleans itself up automatically on the first plan)
- MCR `prefix_filter_lists`: upgrade first, then import existing lists into standalone `megaport_mcr_prefix_filter_list` resources, with a warning not to set `prefix_filter_lists = []` on v1
- MVE `vendor_config` to per-vendor blocks, with a before/after example and a note that the conversion does not replace the MVE

Draft until the removal PRs (#419, #420, #421, #423, #424) merge, since the guide describes their final schema. Depends on all five.


docs/guides/v2_upgrade.md
```diff
@@ -0,0 +1,104 @@
...
+
+1. Upgrade to v2 and remove `prefix_filter_lists` from your `megaport_mcr` configuration.
+2. Add a `megaport_mcr_prefix_filter_list` resource for each existing list.
+3. Import each list using the `mcr_uid:prefix_list_id` format:
+
+   ```shell
+   terraform import megaport_mcr_prefix_filter_list.example "11111111-1111-1111-1111-111111111111:123"
+   ```
+
+Importing preserves the prefix filter list IDs, so any BGP connection filters on your VXCs that reference those IDs keep working. See the `megaport_mcr_prefix_filter_list` resource documentation for details on the import format and finding list IDs.
```
mp1-tomclancy, line 88: Following on from our discussion on #421:

I'm wondering whether an `import` block could be useful here? It would avoid the prefix lists temporarily dropping out of state, before being re-imported again. It also means the whole operation can be reviewed as a single plan. 

I'm pretty sure this would work while the user is on `v1`, e.g.

```terraform
resource "megaport_mcr" "example" {
  # ...
  lifecycle {
    ignore_changes = [prefix_filter_lists]
  }
}

import {
  to = megaport_mcr_prefix_filter_list.allow_private_ipv4
  id = "${megaport_mcr.example.product_uid}:1234"
}

resource "megaport_mcr_prefix_filter_list" "allow_private_ipv4" {
  mcr_id         = megaport_mcr.example.product_uid
  description    = "Allow private IPv4 networks"
  address_family = "IPv4"
  entries        = [ /* ... */ ]
}
```

But I don't know whether it would work if the user has already moved to `v2`.

Phil-Browne: Good call. Checked both halves:

- **On v2 it works fine.** An `import` block goes through the same `ImportResourceState` RPC as `terraform import`, and the resource's `ImportState` (the `mcr_uid:prefix_list_id` parser) is unchanged in v2. Nothing about the block form is v1-specific.
- **On v1 it also works**, as you sketched: the standalone resource and import format already exist there, and the inline/standalone mixing check in v1's `megaport_mcr` only emits warnings, it doesn't block. The one trap is that `ignore_changes = [prefix_filter_lists]` becomes a validation error on v2 once the attribute is gone, so it has to be dropped in the same change as the upgrade.

I've updated the guide (61a2219) to make import blocks the primary method post-upgrade (reviewable as a single plan), kept the CLI as the fallback for older Terraform, and added a short note on your v1 flow for anyone who wants the lists to never leave state, including the `ignore_changes` removal caveat.

mp1-tomclancy (approved):

PR #433, opened 2026-07-22
ESD-1682-fix(v2): warn on silent prefix_filter_lists drop and fix stale MCR migration docs

Phil-Browne:
Split out of #420 so the MCR state-upgrade change gets its own review; #420 stays VXC-only.

Removing the inline `prefix_filter_lists` attribute from `megaport_mcr` was landing silently: anyone still using it would have their lists dropped from Terraform management with no signal that the lists are still live in the backend. This adds a schema-version-0 state upgrader that warns when it sees the removed attribute in prior state, pointing users at the standalone `megaport_mcr_prefix_filter_list` import path.

It also fixes the migration guide, which still told users to add `lifecycle { ignore_changes = [prefix_filter_lists] }` to their MCR even though that attribute no longer exists on the schema (v2 rejects `ignore_changes` for unknown attributes).

**Changes**
- `mcr_resource.go`: bump schema `Version` to 1; add `UpgradeState` that drops inline `prefix_filter_lists` and warns when it was populated.
- `mcr_state_upgrade_test.go`: tests for the warning path and the no-warn (attribute absent) path.
- `docs/index.md` + `templates/index.md.tmpl`: mark the inline approach removed (not deprecated) and drop the stale `ignore_changes` step.

**Note for review:** this deliberately uses a versioned `UpgradeState` rather than the same-version passthrough the rest of the v2 cleanup uses, because emitting a warning needs an upgrader hook. The read-only attribute removals still ride the passthrough; only `prefix_filter_lists` gets the explicit upgrader.

### DIFF
--- a/internal/provider/mcr_state_upgrade_test.go
+++ b/internal/provider/mcr_state_upgrade_test.go
@@ -13,11 +13,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// The V1 to V2 state upgrade needs no upgrader code: the framework's
-// UpgradeResourceState passthrough decodes prior state against the current
-// schema with IgnoreUndefinedAttributes, so removed attributes are dropped.
-// These tests replicate that passthrough to prove real V1 state still
-// decodes against the V2 schema (i.e. no kept attribute changed type).
+// State upgrade coverage for megaport_mcr (provider v1 to v2). The read-only
+// attributes removed in v2 are dropped by the framework's prior-schema decode
+// with IgnoreUndefinedAttributes; the first two tests replicate that decode to
+// prove real V1 state still reads against the V2 schema (no kept attribute
+// changed type). The user-configured prefix_filter_lists attribute instead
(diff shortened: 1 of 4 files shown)


internal/provider/mcr_state_upgrade_test.go
projectsuperba, line 114: Would it make sense to update this test and the other with nil optionals to use the new `runMCRV0Upgrader()` function?

projectsuperba (approved):

internal/provider/mcr_state_upgrade_test.go
Phil-Browne: Submitting a followup PR for this

Phil-Browne: Followed up in #434: both tests (and the nil-optionals one) now go through `runMCRV0Upgrader()`, and the old `decodeV1StateAgainstV2Schema` helper is gone.

PR #435, opened 2026-07-26
docs: clarify Cisco C8000v autonomous mode vendor config

Phil-Browne:
The `megaport_mve` docs say `cloud_init` is required for Cisco C8000v, which steers users ordering a C8000v in autonomous mode into a `400 missing 'cloud_init'` error. There is no `applianceMode` field in the API; the mode is chosen by which credential the order carries. `cloud_init` selects SD-WAN (controller-managed) mode, and `ssh_public_key` without `cloud_init` selects autonomous mode. Verified against the API: key-only orders pass validation on staging and production, and a key-only buy on staging provisioned an autonomous C8000v.

- Clarify the `cloud_init` and `ssh_public_key` descriptions on `vendor_config`.
- Add a C8000v autonomous example.
- Regenerate docs.


internal/provider/mve_resource.go
projectsuperba: Should this still say `admin_ssh_public_key` is required for Cisco?

projectsuperba: Is `manage_locally` technically still required for Cisco?

Phil-Browne: Good catch, it was wrong for all three vendors. `adminSshPublicKey` only appears in four vendor schemas in the megalith spec (F5 BIG-IP, Juniper vSRX, Netskope One Edge, OPNsense), none of which this provider exposes. What Cisco/Fortinet/VMware actually require:

| Spec schema | Required |
|---|---|
| `CiscoC8000` | `oneOf(cloudInit, sshPublicKey)` |
| `CiscoFirewall` (FTDv) | `adminPassword`, `manageLocally` |
| `Fortinet` | `sshPublicKey` |
| `Arista` (VeloCloud/VMware) | `sshPublicKey`, `vcoAddress`, `vcoActivationCode` |

It is also the exact trap this PR is closing: an `adminSshPublicKey`-only C8000v order passes `/v3/networkdesign/validate` but 400s `missing 'cloud_init'` on both the v3 and v4 buy.

Reworded to point at `ssh_public_key` instead.

Phil-Browne: No, only for FTDv. `CiscoFirewall.required` is `[adminPassword, manageLocally]`, and `manageLocally` is not in the `CiscoC8000` schema at all, which matches our own passing `TestAccMegaportMVECisco_Basic` (sets `manage_locally` + `admin_password`, nothing else).

Reworded to match the `admin_password` description, and while in there fixed the three `fmc_*` fields, which claimed to be flatly required for FTDv when the spec has `fmcIpAddress` and `fmcRegistrationKey` required only when `manage_locally` is false, and `fmcNatId` optional in that case and not applicable when true.

projectsuperba (approved):

PR #439, opened 2026-07-29
ESD-1705: Stop IX Read from dropping the resource from state on any API error

Phil-Browne:
`megaport_ix` Read now drops the IX from state only when the API says it's gone. Any other error raises a diagnostic and keeps state.

Before, every `GetIX` error emptied state, and the next plan offered a second IX.

**Changes**

- Classify Read errors with `megaport.IsServiceNotFoundError`.
- Remove a `DECOMMISSIONED` IX from state on a 200.
- Leave a `CANCELLED` IX in state.
	- A cancelled IX stays live and billing until the end of its term.
- Add `MockIXService` and unit tests for the not-found, keep-state, and decommissioned paths.

**Testing**

Build, lint, unit tests, and the Generate gate pass. IX acceptance tests pass against staging and on the harness.

**Scope of change**

Customer-facing yes, a refresh error no longer proposes a duplicate IX. Existing automated coverage yes.


awright-megaport (approved):
all good, thanks @Phil-Browne

PR #445, opened 2026-08-18
ESD-1744: stop a CSP partner-config change from destroying a live VXC

Phil-Browne:
An apply meant only to change a BGP auth key decommissioned two production AWS VXCs. The provider planned a destroy and re-create for an edit to `b_end_partner_config`, Terraform destroyed first, and the re-create failed on a VLAN the destroyed service was still holding.

The edit could never have reached the live VXC anyway: Update does not send a cloud partner config. So the provider now fails the apply and leaves the service alone.

It took two applies to line up, which is why this survived so long. The first apply adds the block, Update ignores it, but `fromAPIVXC` copies it into state, so the apply reports success while the service is untouched. Second apply, state is no longer null, and the changed block plans a replacement.

- No partner-config path appends to `RequiresReplace` any more. The `requiresReplace` field is gone from `vxcEndReconcileInput`. That is the part that fixes the outage.
- The error is raised in `Update`, not at plan time. Terraform does not call `Update` on a replacement, so `-replace`, taint, `replace_triggered_by` and a `service_key` rotation all still rebuild the VXC with the new config. `Update` also runs after unknowns resolve, so a partner config sourced from another resource fails deterministically rather than part way through the apply.
- The gate catches three shapes: removing the block, any change to an end whose live config is a cloud partner, and a partner the provider does not send for that end. All three previously slipped through; the first died late with Terraform's "inconsistent result after apply" error, the other two let `Update` report a success the live service never saw.
- A per-end table now owns which partners the provider sends. A-End takes `vrouter`, `transit` and `a-end`; B-End takes `vrouter` only. `Update` hardcoded that list inline twice and built a B-End `transit` config that `UpdateVXC` rejects with `ErrInvalidVXCBEndPartnerConfig`. That dead branch is gone.
- The errors point at `terraform state pull` for restoring a recorded config, because `state show` masks sensitive values like the vrouter BGP password, and a block restored with a guessed password would pass the gate and clear the live one. The cloud-partner error and both schema descriptions also name the non-destructive fix for a wrongly recorded config: remove the VXC from state, import it again, apply the corrected block.
- Adding a block when state holds none, as after an import, still warns rather than fails. It lands in state, which import depends on. The warning is raised in the `Update` gate, so it fires once per apply instead of on both the plan walk and the apply walk.
- Both schema descriptions promised a delete and re-create. Reworded to say what each end sends on update. Docs regenerated.
- Unit tests split in two: `ModifyPlan` over sixteen cases per end, proving it never proposes a replacement and never fails a plan, and a per-end table driving the `Update` gate over sixteen cases each. Mutation-checked: disabling the gate, loosening the B-End row, and dropping `a-end` from the A-End row each fail multiple subtests.

## What I could not verify

- **The API accepts what the provider refuses to send.** `VxcUpdateRequest` defines `aEndConfig` and `bEndConfig` as the same `oneOf` used at order time, so megalith would take a cloud partner config on update. Nothing here tests whether it honors one, and this PR does not change that the provider never sends it. The error is worded as a provider limit, not an API limit.
- **A cloud partner config still cannot be changed in place.** The error points at `terraform apply -replace`, which rebuilds the service deliberately instead of behind the user's back. That trade is the fix, but it is worth a second opinion.
- **A post-import cloud partner config is recorded, then locked.** The first apply warns and writes it to state. A later correction to that same block hits the gate, because state now holds a cloud partner. The error names the way out (state rm, re-import, re-apply), but a first-class reconcile is a follow-up ticket.
- Whether a CSP auth key can be changed in place at all is a NetAuto question filed separately. The wording can get sharper once that comes back.
- No acceptance test. The gate is provider-side and the unit tests call it directly; acceptance tests are disabled in CI.


projectsuperba (dismissed):
Followup PR suggestion:

1. Verify recorded-only configs in Read — the API returns live per-CSP data in csp_connections (e.g. AWS auth_key). Use it to cross-check (or seed) a partner config that was recorded post-import, instead of trusting the plan verbatim. This removes the warn-then-lock trap: correcting a typo'd recorded value no longer requires state rm + re-import, because state reflects the live VXC rather than an unverified plan.

2. Emit the "recorded in state only" warning once per apply — it currently fires on both the plan and apply walks. Either gate it on the walk or move it out of ModifyPlan.

Phil-Browne:
Second item is fixed in ba18ea6. `ModifyPlan` no longer raises the partner-config diagnostic. The warning moved into `checkPartnerConfigUpdatable`, which `Update` calls once per apply.

I looked for a way to raise the warning on the apply walk only, and the framework doesn't offer one: nothing on `ModifyPlanRequest` says which walk is running (`resource/modify_plan.go` in v1.16.1). This also retires the description's "post-import warning can fire on a replace plan" caveat, since Terraform never calls `Update` on a replacement.

First item is filed as a follow-up rather than built here. Checking it turned up two things:

- The premise holds. The CSP connection response carries enough to rebuild the partner block. Only three attributes aren't recoverable: `aws_config.prefixes`, `azure_config.port_choice`, and `ibm_config.name`.
- Seeding the state doesn't work as-is. The partner-config attributes are `Optional` without `Computed`, so Terraform requires the applied value to match the config. Seeding on refresh creates a diff the provider can't apply. Seeding for a user who wrote no partner block makes Terraform plan a removal, and this PR rejects that plan. Making it work means turning the whole partner-config tree `Optional + Computed`.

The follow-up should verify instead of seed. On a cloud-partner change, read the live CSP connection and compare the planned block against it. If they match, allow the apply: the user is only correcting the record. If they don't, keep the hard error and name the attributes at fault. That covers the typo case too, with no schema change.

There's one trap for whoever picks it up: don't compare against the `csp_connections` state attribute. `fromAPICSPConnection` drops `AmazonASN` and the Azure peers, never sets the IBM name, and puts Oracle's `VirtualCircuitId` into `connection_id`.

projectsuperba (approved):
Code looks fine - the CI checks are stuck in a queued state though, please chase those down to a passing state before merge

PR #446, opened 2026-08-18
ESD-1706: Stop the two MCR ImportState methods re-implementing Read

Phil-Browne:
Terraform runs `Read` right after `ImportState`, so `ImportState` only needs to seed the identifiers. Ten of twelve resources already do that. The two MCR resources (prefix filter list, IPsec add-on) fetched the product and wrote full state, and that copy had drifted from `Read`: 404-only not-found handling, and a `last_updated` stamp on import. `Read` overwrote all of it anyway.

- Trim both `ImportState` methods to parse the import ID and seed the identifiers.
- Add unit tests that run each `ImportState` with a nil client and assert only the identifiers are set.

Importing a gone product now fails with Terraform core's "Cannot import non-existent remote object" instead of the provider's "Error verifying resource during import". The IPsec path also gains `Read`'s wider not-found handling.

Ticket corrections: the IPsec add-on never had `last_updated`, and no import became unused.

**Verified on staging**

- Acceptance tests for both resources pass.
- Live `terraform import` with this build: bogus MCR, decommissioned MCR (IPsec), bogus list ID on a live MCR. All fail with the core error above, no state written.
- Both GET endpoints return 404 for missing IDs. The SDK's 400 "Could not find a service with UID" branch is legacy: megalith mapped that exception to 400 until ENG-34147 (Aug 2026) made it an explicit 404.

Tests pass in CI Acceptance Job

Pre-existing gap, out of scope: prefix list GET on a decommissioned MCR returns 400 "Not an active MCR service", which `Read` does not treat as gone. Tracked in `ENG-35394`


internal/provider/mcr_prefix_filter_list_resource_test.go
```diff
@@ -1188,3 +1192,62 @@ func TestAccMegaportMCRPrefixFilterList_ImportMultipleNoVXCDrift(t *testing.T) {
...
+		t.Errorf("mcr_id = %q, want %q", got.MCRID.ValueString(), "mcr-uid-123")
+	}
+	if got.ID.ValueInt64() != 456 {
+		t.Errorf("id = %d, want 456", got.ID.ValueInt64())
+	}
+	for name, isNull := range map[string]bool{
+		"description":    got.Description.IsNull(),
+		"address_family": got.AddressFamily.IsNull(),
+		"entries":        got.Entries.IsNull(),
+		"last_updated":   got.LastUpdated.IsNull(),
```
nilotpal-barman: Tiny thing, and only because it's about the one attribute whose behaviour actually changes here.

The message says `Read` populates these, which is right for `description`, `address_family` and `entries`. I don't think it holds for `last_updated` though. As far as I can tell it's only assigned in `Create` (line 118) and `Update` (line 214), and `fromAPI` doesn't touch it, so after an import it stays null until the next apply.

That seems fine as behaviour, and arguably better than stamping import time into a field called "last updated". It's `Computed`, and the acceptance tests all ignore it, so nothing breaks. It's just that the message would send the next person looking in `Read` for something that isn't there.

Maybe pull it out of the map with its own check, something like "stays null until the next create or update"? Happy for you to leave it if you'd rather not churn the test.

internal/provider/mcr_ipsec_addon_resource_test.go
```diff
@@ -178,3 +182,55 @@ resource "megaport_mcr_ipsec_addon" "test" {
 }
 `, locationID, mcrName, costCentreName, tunnelCount)
 }
+
+// TestMCRIpsecAddonImportState checks that ImportState seeds the two
+// identifiers and nothing else, leaving the rest to Read. The nil client is the
+// assertion that it makes no API call: any call would panic.
```
nilotpal-barman: Not a change request, I just liked this one. Using the nil client as the assertion that no API call happens is neater than mocking the client out.

Could the comment say why it panics? It works because `client` is a pointer, so any method call on it dereferences nil. If that field ever became a value type with a zero-value method set, the test would quietly stop asserting anything and still pass. One clause would make that dependency visible.

nilotpal-barman (dismissed):
Nice cleanup, and the description made it easy to follow. Deleting a second copy of `Read` that was only ever going to be overwritten is the right call, and I like that you found the drift between the two rather than just noting the duplication.

I checked the things I thought could bite and they all came out fine:

- no imports left stranded by the deletion in either file
- `findIPsecAddOn` still has its other three callers
- both `Read` methods do cover what `ImportState` stopped doing, including the decommissioned-MCR path the old import code was missing
- all six `ImportStateVerify` blocks already ignore `last_updated`, so the one attribute whose behaviour changes can't trip acceptance

Two comments below, both optional, neither holds the merge.

One thing worth noting for whoever reads the release notes rather than the diff: the error on importing something that doesn't exist changes text. Anyone matching on "Error verifying resource during import" would need to update. Low risk given it's an error string on a failure path, and the new one is Terraform's own, but it is the only externally visible change here.

Approving on the basis of reading the diff plus your staging runs and the green acceptance job.

internal/provider/mcr_prefix_filter_list_resource_test.go
Phil-Browne: Checked and you're right: Read never assigns last_updated, only Create and Update do. Pulled it out of the map into its own check with the message "stays null until the next create or update" in 91484c5.

internal/provider/mcr_ipsec_addon_resource_test.go
Phil-Browne: Added the clause in 91484c5. The comment now says the client is a pointer, so any call dereferences nil and panics, which makes the dependency on the field type visible.

nilotpal-barman (approved):
lgtm

PR #447, opened 2026-08-18
ESD-1745: wait for a VXC to decommission before Delete returns

Phil-Browne:
### User-Facing Summary

`terraform destroy` on a `megaport_vxc` now waits for the service to reach `DECOMMISSIONED` before reporting success, bounded by the provider's `wait_time`. Today the destroy returns as soon as the cancel is accepted, and the API returns 200 even when the network termination fails and leaves the service in `CANCELLED`, still live and billed. That case now surfaces as a destroy error naming the VXC and its last status, and the resource stays in state instead of vanishing from Terraform.

- New `waitForVXCDecommission` beside `waitForVXCProvision`: same shape, shared deadline, 30 second ticker, retries transient read errors, treats a not-found read as already gone.
- The wait ends only on `DECOMMISSIONED` or not-found, deliberately not on `CANCELLED`: megalith leaves a service in `CANCELLED` exactly when termination failed and it's still live, which is the case this wait exists to catch.
- `wait_time` docs now mention the destroy wait.

---

### Type of Change

- [ ] 🚀 New Feature
- [ ] ✨ Enhancement
- [x] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### Related Issues

ESD-1745

---

### How to Test

```bash
go test -v -timeout=30m -run 'TestWaitForVXC' ./internal/provider/
```

Five new `TestWaitForVXCDecommission_*` cases: first-poll success, transient read errors then success, not-found, timeout on `CANCELLED`, canceled context. Full suite, `-race -count=2`, and golangci-lint are clean.

The ephemeral acceptance CI ran this branch green: all 8 jobs and 71 tests passed in [run 34406481147](https://github.com/megaport/terraform-provider-acceptance-ci/actions/runs/34406481147). VXC job times match main to within a minute, so the wait adds no measurable teardown time.

---

### Known gaps

- The VLAN is not free when the status reads `DECOMMISSIONED`. Megalith writes that status right after NetAuto's `/terminate` returns 200. The NetAuto team confirmed the VLAN frees only once the deploy-agent finishes, normally one to two minutes later and sometimes more, when NetAuto's own status moves to `Expired`. That status never reaches the customer API, so there is nothing for the provider to poll. A same-VLAN replacement still needs a bounded retry on the create; tracked in ESD-1746.
- Companies with cancellation order approval get a 202 from the cancel and nothing is cancelled. This needs a managed account with `PARTNER_ORDERING_APPROVAL_ENABLED`, a LIVE service, and a non-M2M term. The SDK treats any 2xx as success, so the destroy used to report success immediately; with this wait it now times out after `wait_time` with last status `LIVE`. More honest, still not the right error. Tracked in ESD-1747 (megaportgo #179) and ESD-1748.


Phil-Browne:
Coverage: until today this branch had never run against the ephemeral acceptance stack. Run 34406481147 passed all 8 jobs, 71 tests, on a stack with the 5 second outbox drain from megaport/terraform-provider-acceptance-ci#18. VXC job times match the same harness running main's provider to within a minute, so the wait adds no measurable teardown time.

On the "What I could not verify" section:

- VLAN release: the stack logs show the ordering. NetAuto answers `/secure/ntechnicalservice/terminate` in about 50 ms and megalith writes DECOMMISSIONED right after; the deploy agent tears the network down later on its own loop. DECOMMISSIONED means NetAuto accepted the terminate, not that the VLAN is free. How long the release lags in production is still ESD-1746's question.
- The 202 is real but narrower than "companies with cancellation order approval". It needs a managed account with `PARTNER_ORDERING_APPROVAL_ENABLED`, a LIVE service, and a non-M2M term (`OrderWorkflowApprovalCreationService.orderRequiresApproval`, `LifecycleActionController`). The v3 spec doesn't document the 202 at all.
- "CANCELLED, live, and billed" is half right. `ECommerceService.terminateServiceWithRelatives` terminates billing before it calls the network terminate, so a service stuck in CANCELLED is live but no longer billed. The API still returns 200.

awright-megaport (approved):
all good, thanks @Phil-Browne 👍

PR #455, opened 2026-09-08
ENG-35316: Record the transit partner config on VXC import

### LINKED ISSUE #454: [BUG] Importing a Transit (Megaport Internet) VXC leaves b_end_partner_config null and every later apply fails
**Describe the bug**

I imported a Megaport Internet VXC that was ordered from the Portal. The first `terraform apply` after the import fails:

```
Error: Partner configuration cannot be changed on a live VXC

The provider does not send partner "transit" on the B-End when it updates a
VXC, so this change never reaches the live service. Revert the change, or
replace the VXC with "terraform apply -replace=<resource address>" to build
it with the new configuration. Replacing destroys and rebuilds the service.
```

Every apply after that fails the same way, so I can't move the A-End or change anything else. Dropping `b_end_partner_config` from the configuration doesn't help either. The block is required to create a transit VXC, and `checkPartnerConfigUpdatable` rejects a null plan against a recorded state. I couldn't find any configuration that converges.

**Root cause**

Import and Read never rebuild `b_end_partner_config`, so an imported VXC has it null in state while the configuration has to say `partner = "transit"`. The provider already reads what it would need. `GET /v2/product/{uid}` returns:

```json
"bEnd": { "connectType": "TRANSIT", ... },
"resources": {
  "csp_connection": { "connectType": "TRANSIT", "resource_name": "b_csp_connection", ... }
}
```

That payload ends up in the computed `csp_connections` attribute but not in `b_end_partner_config`. For cloud partners the null-state case is treated as post-import and recorded with a warning. `transit` isn't a cloud partner for `cspPartner()`, so it skips that branch and lands on the `partnerSendableOnUpdate` check, which errors.

**Terraform Provider Version**

- terraform-provider-megaport v1.14.0
- Terraform v1.16.1

**Terraform File and Steps to Reproduce**

The VXC is LIVE and was ordered from the Portal. UIDs are placeholders. The error comes from the provider before any API request.

```hcl
provider "megaport" {
  environment           = "staging"
  accept_purchase_terms = true
}

import {
  to = megaport_vxc.internet
  id = "00000000-0000-0000-0000-00000000vxc1" # existing Megaport Internet VXC
}

resource "megaport_vxc" "internet" {
  product_name         = "Port A to Megaport Internet"
  rate_limit           = 1000
  contract_term_months = 1

  a_end = {
    requested_product_uid = "00000000-0000-0000-0000-0000000port1" # existing Port
    ordered_vlan          = 100
  }

  b_end = {
    requested_product_uid = "00000000-0000-0000-0000-000000transit" # TRANSIT partner port
  }

  b_end_partner_config = {
    partner = "transit"
  }
}
```

1. `terraform plan`

```
  # megaport_vxc.internet will be updated in-place
      + b_end_partner_config = {
          + partner = "transit"
Plan: 1 to import, 0 to add, 1 to change, 0 to destroy.
```

2. `terraform apply`. The import succeeds, the update fails with the error above.

**Expected behaviour**

An imported transit VXC should converge with the configuration that was needed to order it.

**Suggested fix**

Rebuild the block on import, like #452 does for vrouter. `transit` has no configurable fields, so if the B-End `csp_connection` has `connectType == "TRANSIT"` (or `bEnd.connectType == "TRANSIT"`), write

```hcl
b_end_partner_config = { partner = "transit" }
```

into state during import and read. The plan is then empty and nothing is sent.

If that's too much for now, a one-line change in `checkPartnerConfigUpdatable` would at least let transit through the post-import branch the same way cloud partners go:

```go
if statePartnerConfig.IsNull() && (planCSP || planPartner.ValueString() == "transit") {
```

**Additional context**

The same `csp_connection` payload carries the full AWS configuration, so the rebuild approach could later remove the "Partner configuration is recorded in state only" warning for imported cloud-partner VXCs too. I'm leaving that out of this report.

Related: #445 added the gate, #452 rebuilds vrouter partner config on import but doesn't cover transit.

Phil-Browne:
### User-Facing Summary

`terraform import megaport_vxc` on a transit VXC (MVE or port to Megaport Internet) left `b_end_partner_config` null. The first apply then failed: the ESD-1744 check rejects a B-End "transit" config that is not already in state, because the provider cannot send one on update. The only way out was to remove the VXC from state.

The read after an import now records `b_end_partner_config = { partner = "transit" }` when the B-End CSP connection is a transit connection. The transit config carries no settings, so the read has everything the block needs.

- The hook runs on the import read only, keyed off a null `product_name` (Required, so only null after `ImportStatePassthroughID`). A managed refresh leaves the attribute alone, since it is Optional and not Computed and a value written there would plan its own removal.
- The end is matched on `resource_name` `b_csp_connection`, the label megalith puts on the B-End connection. A VXC with more than one match warns and leaves the block for the user.
- `TestMVE_TransitVXC` no longer ignores `b_end_partner_config` on `ImportStateVerify`.
- Schema description updated, docs regenerated.

**Scope of change:** customer facing, yes (import of transit VXCs). `TestVXCRead_RecordsTransitPartnerConfigOnImport` covers the hook over a mock client, and `TestMVE_TransitVXC` covers the import end to end.

Closes #454. Related: #452 rebuilds the vrouter config on the same read and is independent of this change.

---

### Type of Change

- [ ] 🚀 New Feature
- [ ] ✨ Enhancement
- [x] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### Related Issues

ENG-35316, EIP-7497, #454

---

### How to Test

```bash
GOWORK=off go build -v .
GOWORK=off go test -v -timeout=30m -run TestVXCRead_RecordsTransitPartnerConfigOnImport ./internal/provider/
GOWORK=off golangci-lint run --timeout=10m
```

Against staging, with an MVE slot available:

```bash
TF_ACC=1 GOWORK=off go test -v -timeout=60m -run 'TestMVE_TransitVXC$' ./internal/provider/
```


Phil-Browne:
Ran the ephemeral acceptance suite by hand against this branch at f3b096a: https://github.com/megaport/terraform-provider-acceptance-ci/actions/runs/34370278836

All eight jobs passed (ports, MCR, MVE, IX, VXC 1 to 4), 61 tests, about 15 minutes.

One caveat: `TestMVE_TransitVXC` and `TestMVE_TransitVXCAWS` don't carry the `TestAccMegaport` prefix the suite's patterns anchor on, so the run didn't select them. The import fix here is covered by `TestVXCRead_RecordsTransitPartnerConfigOnImport` over the mock client, not by this run. Adding the two transit tests to a VXC job is a follow-up in the acceptance-ci repo.

UbaidSQ (approved):

PR #461, opened 2026-09-10
ENG-35395: Converge the megaport_mcr plan after a prefix filter list migration

### LINKED ISSUE #450: [BUG] Persistent prefix filter drift after migrating existing MCR from inline prefix_filter_lists to megaport_mcr_prefix_filter_list
**Describe the bug**

When an existing MCR was originally created with an inline `prefix_filter_lists` configuration, and we later migrate those lists to the standalone `megaport_mcr_prefix_filter_list` resource (removing the inline attribute and managing filters separately), Terraform continues to report **drift on the prefix filter list**.

This only affects MCRs that had prefix filters configured **before** the migration to the separate resource.

**Control tests (same provider/module, not affected):**
1. Existing MCR that never had a prefix filter list → **no drift**
2. New MCR deployed with `megaport_mcr_prefix_filter_list` from day one → **no drift**

So the issue appears specific to the **migration path** for pre-existing inline prefix filter lists, not to the standalone resource itself.

**Terraform and Provider Version**
- Terraform: `v1.5.7`
- Provider: `megaport/megaport` `1.13.1`

**Terraform File and Steps to Reproduce**
1. Create/apply an MCR with **inline** prefix filters:

```hcl
resource "megaport_mcr" "example" {
  product_name         = "example-mcr"
  port_speed           = 1000
  location_id          = <location_id>
  contract_term_months = 12
  asn                  = 65000

  prefix_filter_lists = [{
    description    = "example-filter"
    address_family = "IPv4"
    entries = [{
      action = "permit"
      prefix = "10.0.0.0/24"
      ge     = 24
      le     = 24
    }]
  }]
}


2. Migrate to the standalone resource (as documented): remove prefix_filter_lists from megaport_mcr, add lifecycle { ignore_changes = [prefix_filter_lists] }, and create megaport_mcr_prefix_filter_list:

```hcl

resource "megaport_mcr" "example" {
  product_name         = "example-mcr"
  port_speed           = 1000
  location_id          = <location_id>
  contract_term_months = 12
  asn                  = 65000
}

resource "megaport_mcr_prefix_filter_list" "example" {
  mcr_id         = megaport_mcr.example.product_uid
  description    = "example-filter"
  address_family = "IPv4"
  entries = [{
    action = "permit"
    prefix = "10.0.0.0/24"
    ge     = 24
    le     = 24
  }]
}

3. Import/adopt the existing filter list into state as needed, then run terraform plan repeatedly.
4. Observed: ongoing drift related to the prefix filter list for this migrated MCR.
5. Not observed: same drift for MCRs that never had inline filters, or for brand-new MCRs created with the standalone resource only.

Additional context 

1. Provider docs note that ignore_changes = [prefix_filter_lists] - https://registry.terraform.io/providers/megaport/megaport/latest/docs

### PR DESCRIPTION
**Pull request by Phil-Browne**
### User-Facing Summary

Migrating an MCR off the deprecated inline `prefix_filter_lists` attribute left the plan churning forever. Once the attribute came out of the configuration, every `terraform plan` proposed an in-place update of the same twelve computed attributes, apply wrote the same values back, and the next plan repeated. The plan never named `prefix_filter_lists`, so there was nothing to point at.

Removing an Optional+Computed nested attribute makes Terraform core propose `null` for it. The framework reads that as a change and marks every Computed attribute with a null config value as unknown. It does that before any plan modifier runs, so no modifier on `prefix_filter_lists` can prevent it.

- Give `megaport_mcr` a `ModifyPlan` method that restores prior state for those attributes, but only when the plan holds no other change.
- A real change leaves every unknown alone, so an update still refreshes `last_updated` and the API-sourced fields.

The documented `lifecycle { ignore_changes = [prefix_filter_lists] }` workaround still works, and becomes optional.

**Scope of change:** customer-facing (plan behavior for every `megaport_mcr` user). Covered by three new unit tests and one new acceptance test; acceptance tests do not run in CI.

**What I could not verify:** the two guard unit tests (`KeepsUnknownsOnRealChange`, `SkipsCreateAndDestroy`) pass with or without the fix, since they assert the plan is left alone. Only `ConvergesMigrationDrift` and the acceptance test fail on `main`.

**Other resources:** `megaport_port` and `megaport_lag_port` don't have the trigger. Their only nested attribute, `resources`, is Computed-only, so it can't be removed from a configuration. `megaport_mve` does: `vnics` is Optional+Computed with a Required child, the same shape as `prefix_filter_lists`, and seven of its computed attributes carry no modifier. ENG-35420 covers it, and lifts this loop into a shared helper rather than copying it.

---

### Type of Change

- [ ] 🚀 New Feature
- [ ] ✨ Enhancement
- [x] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### Related Issues

Fixes #450. Jira: ENG-35395.

---

### How to Test

Unit tests, no credentials needed:

```
go test -v -run TestMCRResourceModifyPlan ./internal/provider/
```

Acceptance test, needs credentials:

```
TF_ACC=1 go test -v -timeout=40m -run TestAccMegaportMCR_InlinePrefixListMigration ./internal/provider/
```

By hand: apply an MCR with an inline `prefix_filter_lists` block, remove the block, then plan. Before this change the plan proposes an update of twelve computed attributes and never settles. After it, the plan is empty.

---

### DIFF
--- a/internal/provider/mcr_resource.go
+++ b/internal/provider/mcr_resource.go
@@ -25,6 +25,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
 	megaport "github.com/megaport/megaportgo"
 )
 
@@ -33,6 +34,7 @@ var (
 	_ resource.Resource                = &mcrResource{}
 	_ resource.ResourceWithConfigure   = &mcrResource{}
 	_ resource.ResourceWithImportState = &mcrResource{}
+	_ resource.ResourceWithModifyPlan  = &mcrResource{}
 
 	mcrPrefixFilterListModelAttributes = map[string]attr.Type{
 		"id":             types.Int64Type,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Phil-Browne** [comment]
Copilot's second pass asks for a closer look because the acceptance test is not exercised in CI. That is true today, so here is the evidence from running it by hand.

`TestAccMegaportMCR_InlinePrefixListMigration` passes against staging, all five steps, 96.8s. `TestAccMegaportMCR_Basic` is unchanged and still passes (163s), which covers the inline attribute path.

Mutation check: with `ModifyPlan` stubbed to return immediately, step 2 fails and reproduces the reported plan exactly.

```
Plan: 0 to add, 1 to change, 0 to destroy.
~ aggregation_id      = 0 -> (known after apply)
~ attribute_tags      = {} -> (known after apply)
~ cancelable          = true -> (known after apply)
~ company_name        = "Megaport Lab" -> (known after apply)
~ contract_end_date   = "2027-09-10 10:27:06 -0700 PDT" -> (known after apply)
~ contract_start_date = "2026-09-10 10:27:06 -0700 PDT" -> (known after apply)
~ lag_id              = 0 -> (known after apply)
~ lag_primary         = false -> (known after apply)
~ last_updated        = "Thursday, 10-Sep-26 10:27:29 PDT" -> (known after apply)
~ live_date           = "2026-09-10 10:27:06 -0700 PDT" -> (known after apply)
~ provisioning_status = "LIVE" -> (known after apply)
~ terminate_date      = null -> (known after apply)
```

No `acceptance` label on this PR yet: the workflow that consumes it is still open in #456, and a `pull_request_target` workflow runs from the base branch, so the label does nothing on `main` today.

awright-megaport (approved):
awesome, thanks @Phil-Browne 👍

PR #463, opened 2026-09-10
ENG-35346: Pin the harness ref when dispatching acceptance runs

Phil-Browne:
The `acceptance` label on #387 failed at the dispatch step: `gh workflow run` with no `--ref` looks up the harness default branch over GraphQL, and that field needs `contents:read`. The App token deliberately carries only `actions:write` and `metadata:read`.

- Pass `--ref main` to `gh workflow run` so the lookup never happens.
- No token permission change. Every other call on the path (`ResolveWorkflow`, the dispatch POST) is REST under `actions`, which the earlier cancel step already proved works.

**Scope of change:** not customer-facing. No automated coverage; `pull_request_target` reads this workflow from `main`, so the check is to merge and re-label #387.

Failing run: https://github.com/megaport/terraform-provider-megaport/actions/runs/34541211482

### DIFF
--- a/.github/workflows/acceptance-label.yaml
+++ b/.github/workflows/acceptance-label.yaml
@@ -82,7 +82,9 @@ jobs:
         env:
           GH_TOKEN: ${{ steps.harness-token.outputs.token }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
-        run: gh workflow run acceptance.yml -R "$HARNESS" -f provider_ref="$HEAD_SHA" -f pr_number="$PR"
+        # Without --ref, gh resolves the harness default branch over GraphQL,
+        # which needs contents:read that this token deliberately lacks.
+        run: gh workflow run acceptance.yml -R "$HARNESS" --ref main -f provider_ref="$HEAD_SHA" -f pr_number="$PR"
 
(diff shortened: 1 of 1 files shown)


awright-megaport (approved):
all good 👍

Phil-Browne:
Acceptance dispatched by hand with the fixed command (the label path reads this workflow from `main`, so it cannot test itself):

- [Run 34543315918](https://github.com/megaport/terraform-provider-acceptance-ci/actions/runs/34543315918): 7 of 8 green. VXC 3 had failed on the [first run](https://github.com/megaport/terraform-provider-acceptance-ci/actions/runs/34542232472) with a megalith 500 on an MCR read (`trace_id 333ce5fcdf5b5febbe361283d0297fbb`) and passed on the rerun.
- MCR fails the test-count gate on both runs (15 ran, expected 16). The harness bumped the count to 16 for `TestAccMegaportMCR_InlinePrefixListMigration`, which is in #461 and not yet on `main`. Every run against `main`, including the nightly, fails MCR until #461 merges. Unrelated to this change.

Both runs landed check runs on this head and were titled `PR #463 at <sha>`, so the dispatch and the report-back paths both work.

awright-megaport (approved):

Issue #465, opened 2026-09-14
[BUG] docs: vrouter partner_config added after import IS sent and applied in-place — schema descriptions say otherwise


maksim-humbility:
**Describe the bug**
The schema descriptions for `a_end_partner_config` / `b_end_partner_config` on `megaport_vxc` state that a partner configuration added after an import is recorded in state but not sent, and that changing a recorded configuration requires removing the VXC from state and importing again.

For **vrouter** configurations this does not match observed behavior — and the observed behavior is better than documented:

1. We imported a live MCR-to-AWS VXC (hosted VIF) whose state therefore had `a_end_partner_config = null`.
2. We then declared a full `a_end_partner_config` with `partner = "vrouter"` and a `vrouter_config` transcribed 1:1 from the live portal configuration (interfaces, BGP connection, prefix filter lists referenced by name).
3. `terraform plan` showed the block as an in-place update (no replace). `terraform apply` **sent it** — the vRouter configuration was applied on the Megaport side — and subsequent plans converged (no perpetual diff).

This makes brownfield adoption of existing VXCs possible without destroy/recreate, which is exactly what a customer with a live estate needs. Please confirm this is intended behavior for vrouter configs and update the schema descriptions accordingly (the "recorded but not sent" wording appears to apply only to cloud-partner configs). Note the password caveat filed separately — an omitted `password` clears the live BGP session key when the block is sent.

**Terraform and Provider Version**
terraform-provider-megaport 1.14.0, OpenTofu 1.x.

**Terraform File and Steps to Reproduce**
As above: import a live MCR VXC by UID, then add an `a_end_partner_config` vrouter block matching the live config and apply. Product UIDs available on request.

**Additional context**
Related docs: docs/resources/vxc.md, `a_end_partner_config` / `b_end_partner_config` attribute descriptions.

Phil-Browne:
Thanks for the detailed write-up, and yes, this is intended for vrouter configurations.

The provider sends a vrouter partner configuration on update, including one added to a VXC whose imported state had none. `Update` compares the planned and prior partner config before it normalizes the null state an import leaves behind, so a null-to-vrouter change is a real change and goes to the API. The "recorded but not sent" wording applies to cloud partner configurations only.

[...]

PR #469, opened 2026-09-15
ENG-34866: State when vrouter config is added after import is sent

Phil-Browne:
This PR states outright in both `megaport_vxc` partner_config descriptions that a vrouter configuration added or changed after an import is sent and applied in place.

Issue was reported in #465: the wording read as if any partner config added after import was recorded and never sent.

**Changes**

- Add one sentence to the `a_end_partner_config` description.
- Add the same sentence to the `b_end_partner_config` description.
- Name the provider as the subject of the next sentence.
	- The new sentence otherwise left "It does not send" pointing at the vrouter config.
- Regenerate `docs/resources/vxc.md`.

**Testing**

Build, lint, unit tests, and the Generate gate all pass. The described behavior was read off `Update` in `vxc_resource.go`, where partner config change detection runs before null state is normalized.

**Scope of change**

Customer-facing yes, docs only. Existing automated coverage yes, no behavior changed.


nilotpal-barman (approved):
lgtm.

PR #470, opened 2026-09-15
ENG-35569: Order the Oracle VXC test against the port its key resolves to

Phil-Browne:
### User-Facing Summary

`TestAccMegaportOracleVXCWithProductUID` now orders its b-end against the partner port the Oracle key resolves to, not a fixed port at location 147.

That port is US Gov Cloud (oc2) and every key in the pool is oc1, so NetAuto looked the key up in the wrong realm.

**Changes**

- Return `cspPickResult{Key, PartnerPortUID}` from `pickOracleVirtualCircuitID`.
  - The lookup already ran and discarded the port.
- Drop the `data.megaport_partner` block pinned to location 147.
- Set `b_end.requested_product_uid` to the resolved port, as the Azure test does.
- Update `TestAccMegaportOracleVXC_Basic` for the new return type only.

**Testing**

`go vet`, unit tests, and golangci-lint pass. Ran both Oracle tests against staging: `WithProductUID` passed create, checks, and destroy. Acceptance stack not run, so fire the `acceptance` label after review.

`OracleVXC_Basic` failed on destroy with a 500 in that run. This PR does not change it: its VXC comes up with no `csp_id`, which needs its own ticket.

**Scope of change**

Not customer-facing. Test-only. Existing automated coverage: yes, these two acceptance tests.

---

### Type of Change

- [x] 🐛 Bug Fix

---

### How to Test

Run on the acceptance stack with the pinned csp-mock image and no `oracle.json`
override:

```bash
TF_ACC=1 go test -v -timeout 30m ./internal/provider/ \
  -run 'TestAccMegaportOracleVXC(WithProductUID|_Basic)$'
```

The `WithProductUID` log shows no `location_id = 147`, and destroy completes.


tuantran-rgb (approved):
LGTM.

PR #474, opened 2026-09-16
ENG-35628: Do not replace a VXC when service_key is added after import

Phil-Browne:
### User-Facing Summary

---

Adding `service_key` to an imported `megaport_vxc` no longer replaces the VXC. Any other change to the key still replaces it.

The API never returns the key, so adding it back to the configuration after an import destroyed the live VXC.

A VXC imported with an earlier provider version still plans a replace when the key is added. Remove it from state and import it again first.

**Changes**

- Swap `RequiresReplace` for `RequiresReplaceIf` on `service_key`.
- Mark an import in private state.
  - This tells an imported VXC apart from one that never had a key.
- Copy the planned key into state on update.
  - The API read cannot supply it.
- Warn that the post-import apply records the key and does not send it.
- Reword the attribute description and regenerate docs.
- Add unit tests and an acceptance test for the import path.

### Type of Change

- [ ] 🚀 New Feature
- [ ] ✨ Enhancement
- [x] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### Related Issues

ENG-35628

---

### How to Test

**Testing**

Unit tests and lint pass. On staging I imported a live VXC, applied a configuration with the key, and the UID stayed the same.

**Scope of change**

Customer-facing: yes. Existing automated coverage: no, tests added here.


internal/provider/vxc_resource_utils.go
```diff
@@ -9,12 +9,21 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 	megaport "github.com/megaport/megaportgo"
 )
 
+// requiresReplaceServiceKey replaces the VXC only when state already holds a
+// service key. The API never returns the key, so an imported VXC has it null
+// and the first apply records the configured value in place.
+func requiresReplaceServiceKey(_ context.Context, req planmodifier.StringRequest, resp *stringplanmodifier.RequiresReplaceIfFuncResponse) {
+	resp.RequiresReplace = !req.StateValue.IsNull()
```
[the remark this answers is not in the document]
Phil-Browne: Good catch. Fixed in b389260: `Read` now sets a private-state flag only on the read that follows an import (the same signal the vrouter/transit partner config repairs already use), and the plan modifier only skips the replace when that flag is set. An ordinary VXC created without a key still replaces when one is added, and `Update` clears the flag once the key is recorded. Re-ran `TestAccMegaportVXC_ImportDrift_ServiceKey` against staging to confirm the genuine post-import path still works (PASS, 276.85s).

internal/provider/vxc_resource_utils.go
```diff
@@ -9,12 +9,36 @@ import (
...
+// must still replace so the key actually reaches the API.
+func requiresReplaceServiceKey(ctx context.Context, req planmodifier.StringRequest, resp *stringplanmodifier.RequiresReplaceIfFuncResponse) {
+	if !req.StateValue.IsNull() {
+		resp.RequiresReplace = true
+		return
+	}
+
+	imported, diags := req.Private.GetKey(ctx, serviceKeyImportedPrivateKey)
+	resp.Diagnostics.Append(diags...)
+	resp.RequiresReplace = len(imported) == 0
```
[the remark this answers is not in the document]
Phil-Browne: Agreed the flag proves import, not which key the live VXC was ordered with. That gap cannot be closed from the provider: `service_key` is order-time only and absent from every VXC read shape in the megalith spec (`ProductVxcDto`, `VxcEndDto`), so there is nothing to validate the planned key against. Reading it back is also out of scope for this ticket.

That leaves a policy choice between two failure modes. Recording an unverifiable key in place is a benign miss: the live VXC keeps running unchanged. Replacing on every null-to-value change is the destructive one this ticket exists to fix: it tears down a live imported VXC and burns a single-use key. Keeping the in-place path.

`244090b` states the limitation in the attribute description so users know that apply records the key rather than sends it.

internal/provider/vxc_resource.go
```diff
@@ -2764,6 +2778,13 @@ func (r *vxcResource) Update(ctx context.Context, req resource.UpdateRequest, re
 	apiDiags := state.fromAPIVXC(ctx, vxc, tags, &plan)
 	state.LastUpdated = types.StringValue(time.Now().Format(time.RFC850))
 	state.PromoCode = plan.PromoCode
+	state.ServiceKey = plan.ServiceKey
+	if !plan.ServiceKey.IsNull() && resp.Private != nil {
```
nilotpal-barman, line 2826: With the `!plan.ServiceKey.IsNull()` check, I think an imported VXC with no key would keep the flag forever. If a key was added later, we'd record it but not send it, and because the API never returns the key, we'd never see the difference. Would it be safer to clear the flag on the first Update after import? The downside is that `import → unrelated apply → add key` would then plan a replace. Both seem reasonable to me. What do you think?

internal/provider/vxc_resource.go
```diff
@@ -2106,6 +2110,16 @@ func (r *vxcResource) Read(ctx context.Context, req resource.ReadRequest, resp *
...
 		resp.Diagnostics.Append(state.fillTransitPartnerConfigOnImport(ctx, vxc)...)
+
+		// service_key is order-time only, so the API never returns it: this
+		// flag is the only way the plan modifier can tell a null key that
+		// might still be live on the imported VXC from one that never
+		// existed. Update clears it once the key is recorded. Terraform
+		// always supplies Private; nil only happens when a test calls Read
+		// directly without going through the protocol server.
+		if resp.Private != nil {
+			resp.Diagnostics.Append(resp.Private.SetKey(ctx, serviceKeyImportedPrivateKey, []byte("true"))...)
```
nilotpal-barman, line 2152: Since this only runs on the post-import read, I don't think existing state would pick up the flag. Anyone already hitting this would upgrade and still see the destroy plan, and would need to do a state rm and re-import. Might be worth adding that to the `service_key` description so people know?

nilotpal-barman (dismissed):
Thanks for this @Phil-Browne, LGTM. Added two suggestions inline. Kindly have a look and see what you think 👍 .

internal/provider/vxc_resource.go
Phil-Browne: Thanks for catching this @nilotpal-barman 

My understanding is that neither VXC update endpoint accepts a service key in the megalith spec (VxcUpdateRequest and ServiceUpdateRequest), and the Java update path never reads one. It appears that Megalith keeps the order-time key on the stored order and uses its max speed to cap later rate limit changes, so the key still matters, but the client cannot set it after ordering. 

If this is accurate, the flag meaning "key history unknown" would stay true until a key is recorded. Please let me know if any of this is off re megalith-side. 

I'm leaning toward keeping it, but with an added warning in https://github.com/megaport/terraform-provider-megaport/pull/474/commits/91951419708ad8befcf93fd3bc71436137d47674 on apply that records the key, matching the cloud partner config warning. 

How does that sound to you?

internal/provider/vxc_resource.go
Phil-Browne: Good catch on this, state from an earlier version has no flag, so adding the key after upgrade still plans a replace. 

Commit [9195141](https://github.com/megaport/terraform-provider-megaport/pull/474/commits/91951419708ad8befcf93fd3bc71436137d47674) adds that to the service_key description with the state rm and re-import steps, and the PR body now carries the same note for the release.

internal/provider/vxc_resource.go
nilotpal-barman: All good here 👍

nilotpal-barman (approved):
LGTM

PR #478, opened 2026-09-18
ENG-35633: Record AWS prefixes on VXC import

Phil-Browne:
The VXC import now records prefixes on an AWS partner configuration instead of leaving them null. ENG-35629 added the SDK field this consumes.

**Changes**

- Bump megaportgo to v1.20.0 for `CSPConnectionAWS.Prefixes`
- Map `conn.Prefixes` onto `aws_config.prefixes` on import
- Drop prefixes from the "settings left null" import warning
- Revert the `prefixes` description's stale import-null sentence

**Testing**

`go build`, `go test ./internal/provider/`, `golangci-lint run`, and `go generate ./...` (diff-clean) all pass. Added a no-prefixes subtest alongside the flipped ENG-35626 one to cover both AC branches.

**Scope of change**

Customer-facing: yes, an import-time behavior change. Existing automated coverage: yes, unit tests only, per the-staging-csp-mock-does-not-echo-the-aws-prefixes-key gotcha.


internal/provider/vxc_resource_utils.go
```diff
@@ -950,11 +950,12 @@ func (orm *vxcResourceModel) fillBEndPartnerConfigOnImport(ctx context.Context,
 			ConnectType:    stringOrNull(conn.ConnectType),
 			Type:           stringOrNull(conn.Type),
 			OwnerAccount:   stringOrNull(conn.OwnerAccount),
+			Prefixes:       stringOrNull(string(conn.Prefixes)),
```
nilotpal-barman, line 953: Please correct me if I’m wrong here, but I think an imported VIF with prefixes left out of the config would now fail on the next apply. State has a value that the config doesn’t, so `onlyFillsNulls` returns false and `checkPartnerConfigUpdatable` errors instead of just warning.

Might be worth calling that out in the import warning? The `peer_type` note at `vxc_resource.go:1210` says a missing value will show as a change on the next plan, but that’s not quite strong enough here, since a cloud partner change can’t actually be sent.

internal/provider/vxc_resource_utils.go
```diff
@@ -950,11 +950,12 @@ func (orm *vxcResourceModel) fillBEndPartnerConfigOnImport(ctx context.Context,
...
 			ConnectionName: stringOrNull(conn.Name),
 		})
 		diags.AddWarning(
 			summary,
-			"The import leaves aws_config.asn, aws_config.amazon_asn, aws_config.auth_key, aws_config.customer_ip_address, aws_config.amazon_ip_address, and aws_config.prefixes null in b_end_partner_config. AWS assigns those values when the order leaves them out, and the read cannot tell an assigned value from one the configuration set. "+recordsOnNextApply,
+			"The import leaves aws_config.asn, aws_config.amazon_asn, aws_config.auth_key, aws_config.customer_ip_address, and aws_config.amazon_ip_address null in b_end_partner_config. AWS assigns those values when the order leaves them out, and the read cannot tell an assigned value from one the configuration set. "+recordsOnNextApply,
```
nilotpal-barman: Does the API actually return prefixes for a VIF? Your note about the staging mock made me wonder. If it does, might be worth adding a line here saying the import records whatever the API reports.

nilotpal-barman (changes requested):
Thanks for this @Phil-Browne. 
I have added 2 review comments. Kindly have a look.

internal/provider/vxc_resource_utils.go
Phil-Browne: Correct. If the API returns prefixes, the import records it, and a config without it fails the next apply. I've extended the import warning a bit in https://github.com/megaport/terraform-provider-megaport/pull/478/commits/3ca1c2ab2c9ff45d61dd3ec138e3b75b7327d12a thanks!

internal/provider/vxc_resource_utils.go
Phil-Browne: Yes, for public VIFs. Private VIFs don't have prefixes, and the staging mock never returns it. I added a line to the warning saying the import records what the API reports in https://github.com/megaport/terraform-provider-megaport/pull/478/commits/3ca1c2ab2c9ff45d61dd3ec138e3b75b7327d12a thanks!

nilotpal-barman (approved):
Thanks @Phil-Browne. LGTM

PR #479, opened 2026-09-18
ENG-35720: Fix the deprecated VXC A-End partner config interface decode

Phil-Browne:
Fixes the deprecated `partner_a_end_config` block so its interfaces decode instead of failing every apply. The decode used the vrouter model's wider struct, which no A-End interface object can match field-for-field.

**Changes**

- Add `vxcPartnerConfigAEndInterfaceModel`, a 5-field A-End decode target
- Add `aEndBgpConnectionConfigModel` for the nested BGP decode
  - The A-End `bgp_connections` object has no `peer_type`
- Drop the `ip_mtu` read from the A-End interface decode
  - The A-End schema never exposed `ip_mtu`
- Correct the attr-map comment's claim about unmatched fields
- Add `TestCreateAEndPartnerConfig_Interfaces`, covering both decode paths

**Testing**

Unit test fails on `origin/main`, passes on this branch. `go build`, `golangci-lint`, and `go generate` are all clean. No acceptance test: the fix targets a decode helper, not a full apply path.

**Scope of change**

Customer-facing: yes. Existing automated coverage: no, this PR adds the first test for the block.


nilotpal-barman (approved):
LGTM

PR #482, opened 2026-09-21
ENG-35723: Grow a LAG in place when lag_count increases

Phil-Browne:
### User-Facing Summary

Raising `lag_count` now adds ports to the LAG the customer already has, instead of cancelling every member and ordering a new LAG.

**Changes**

- Order the missing ports against the LAG's aggregation id.
- Count the ports to order from a live read, not from state.
    - A retry after a failed provisioning wait then orders nothing.
    - A live count below state fails the apply instead of over-ordering.
- Replace the LAG only when `lag_count` goes down.
- Mark `lag_port_uids` unknown when the count goes up.
- Send the promo code and resource tags on the added ports.
- Name the ordered ports when a grow fails before the state write.
- Describe both directions in the `lag_count` docs.

---

### Type of Change

- [ ] 🚀 New Feature
- [x] ✨ Enhancement
- [ ] 🐛 Bug Fix
- [ ] 📚 Documentation Update
- [ ] 🏗️ Chore / Other

---

### Related Issues

ENG-35723.

**Blocked on** megaport/megaportgo#192. `go.mod` pins that PR's head, and it has to move to a tagged release before merge. This PR stays a draft until then.

---

### How to Test

Unit tests cover the plan decision and the order body against a stub server, so both run in CI. `TestAccMegaportLAGPort_GrowLagCount` grows a two-port LAG on staging and passes. The shrink step is plan-only, so no apply exercises the cancel.

**Scope of change**

Customer-facing yes, existing automated coverage yes.


nilotpal-barman (commented):
Thanks for this @Phil-Browne.

One question on the retry. I don't think a new `LAG` port gets an aggregation id until it is configured. In prod, ports that are still `NEW` haven't been assigned to a physical interface yet, so they don't have an aggregation id. `GetPort` only counts ports that share the aggregation id, so I don't think it would see them.

So if the new ports are still `NEW` when `wait_time` runs out, at a pre-order location for example, the apply fails. I think the next apply would then see the old count and order the ports again. `BuyPort` also drops the new UIDs when it gives up waiting, so there's nothing left to find them by. What do you think?

Phil-Browne:
> Thanks for this @Phil-Browne.
> 
> One question on the retry. I don't think a new `LAG` port gets an aggregation id until it is configured. In prod, ports that are still `NEW` haven't been assigned to a physical interface yet, so they don't have an aggregation id. `GetPort` only counts ports that share the aggregation id, so I don't think it would see them.
> 
> So if the new ports are still `NEW` when `wait_time` runs out, at a pre-order location for example, the apply fails. I think the next apply would then see the old count and order the ports again. `BuyPort` also drops the new UIDs when it gives up waiting, so there's nothing left to find them by. What do you think?

I've dropped the retry claim, and the error now warns that applying again before the new ports provision orders more in https://github.com/megaport/terraform-provider-megaport/pull/482/commits/4b462a2d7872d4a29e7c7e77c5c0f0d95513e412. Submitting a follow-up ticket for the SDK shortly.

nilotpal-barman (approved):
Thanks @Phil-Browne. LGTM

PR #484, opened 2026-09-22
ENG-35812: Reject VXC partner config blocks that do not match partner

Phil-Browne:
This PR adds safeguards for VXC Partner Configs with nested blocks that its `partner` value doesn't use. With this fix, the provider now fails the plan.

Prior to this PR, the order would drop that block and the state builders would null it, causing the apply to fail after the VXC had been ordered.

**Changes**

- Add `ValidateConfig` to `megaport_vxc`, covering both ends
- Report every mismatched block, not just the first
- Skip an unknown block and an unknown `partner`
- State the rule in both partner config descriptions, regenerate docs

**Testing**

Eight unit tests drive `ValidateConfig` through the resource. One acceptance step fails a real plan and places no order. Acceptance tests passing against staging. Acceptance CI is currently awaiting version re-pin merge.

**Scope of change**

Customer-facing: yes, a config that Terraform accepted before now fails the plan. Existing automated coverage: yes.


UbaidSQ (approved):
Looks good

